PR #24, opened 2019-09-28
Add multi-day time filters

pxpeterxu:
In GitLab by @michaellinden on Aug 14, 2019, 02:23

[Background] Coursetable currently only supports sorting by exactly one day. It would be desirable to be able to sort by multiple days to find classes to fit a schedule.

[Change] Instead of storing the day_of_week property as a string, we convert it to an array, and convert all the filtering code to support this new array structure. Sorting may now be done by multiple days at a time, and the clear button still works.

[Test] Tested locally to confirm everything works.

### DIFF
--- a/web/includes/ProjectCommon.php
+++ b/web/includes/ProjectCommon.php
@@ -1,5 +1,5 @@
 <?php
-require_once __DIR__ . '/../../web/libs/cas/CAS/Autoload.php'; // change path as needed
+require_once __DIR__ . '/../../vendor/autoload.php'; // change path as needed
 
 $filePath = dirname(__DIR__);
 define('FILE_PATH', $filePath);
@@ -81,17 +81,16 @@ class ProjectCommon
 
     public static function casAuthenticate($force = true)
     {
-        // if ($force) {
-        //     phpCAS::forceAuthentication();
-        // }
+        if ($force) {
+            phpCAS::forceAuthentication();
+        }
 
-        // $netId = null;
-        // if (phpCAS::isAuthenticated()) {
-        //     $netId = phpCAS::getUser();
-        // }
+        $netId = null;
+        if (phpCAS::isAuthenticated()) {
+            $netId = phpCAS::getUser();
+        }
 
-        // return $netId;
-        return 'mdl74';
+        return $netId;
     }
 
     public static function casLogout()
--- a/web/includes/ProjectCommon.php
+++ b/web/includes/ProjectCommon.php
@@ -1,5 +1,5 @@
 <?php
-require_once __DIR__ . '/../../vendor/autoload.php'; // change path as needed
+require_once __DIR__ . '/../../web/libs/cas/CAS/Autoload.php'; // change path as needed
 
 $filePath = dirname(__DIR__);
 define('FILE_PATH', $filePath);
@@ -81,16 +81,17 @@ class ProjectCommon
 
     public static function casAuthenticate($force = true)
     {
-        if ($force) {
-            phpCAS::forceAuthentication();
-        }
+        // if ($force) {
+        //     phpCAS::forceAuthentication();
+        // }
 
-        $netId = null;
-        if (phpCAS::isAuthenticated()) {
-            $netId = phpCAS::getUser();
-        }
+        // $netId = null;
+        // if (phpCAS::isAuthenticated()) {
+        //     $netId = phpCAS::getUser();
+        // }
 
-        return $netId;
+        // return $netId;
+        return 'mdl74';
     }
 
     public static function casLogout()
--- a/web/js/filterui.js
+++ b/web/js/filterui.js
@@ -665,10 +665,18 @@ export function TimeFilterUi() {
 
     $weekDayButtons.off('click').on('click', event => {
       const buttonText = $(event.delegateTarget).text();
-      const changedWeekDay = currentFilter.day_of_week !== buttonText;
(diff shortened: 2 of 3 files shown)


pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 00:13

added 18 commits

<ul><li>fbdc8ead...867e2158 - 15 commits from branch <code>master</code></li><li>54b4b7d0 - prelimanary changes</li><li>f8f7ea0e - undo auth ovveride</li><li>ab48bd7c - undo local path change</li></ul>

[Compare with previous version](/coursetable/coursetable/merge_requests/4/diffs?diff_id=55037344&start_sha=fbdc8eadbf881541985dfc6bb02cafe5d110f7a5)

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 00:27

added 1 commit

<ul><li>268e95bb - Rebased package-lock on master</li></ul>

[Compare with previous version](/coursetable/coursetable/merge_requests/4/diffs?diff_id=55037675&start_sha=ab48bd7cb3f0d178751e99e52ccc0416d0ad97a8)

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 01:46

added 6 commits

<ul><li>268e95bb...216f9275 - 3 commits from branch <code>master</code></li><li>56d3a2da - prelimanary changes</li><li>e798ada3 - undo auth ovveride</li><li>d4d27269 - undo local path change</li></ul>

[Compare with previous version](/coursetable/coursetable/merge_requests/4/diffs?diff_id=55039277&start_sha=268e95bb8d5d8a8beca1ed1d6a960f23b18393ff)

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:13

`dayOfWeek` here could probably now be renamed `daysOfWeek`

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:14

Now that it's an array, it may be nice to rename this `days_of_week` so it's clear. This will require changing code in `filterManager.registerFilterGenerator('times', ...` and in all of `filterui.js`

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:20

Here's the code that errors. It's because we're trying to use `dayOfWeek`, which is now an array, as the index of the array to fetch! This fails to retrieve anything, and thus `timesOnDay` is undefined, and trying to call `undefined.length` results in an error.

To fix this, we'd probably want to loop through daysOfWeek.

Let's assume we renamed `dayOfWeek` to `daysOfWeek` per the suggestion above. The for loop could look something like:

```
for (const dayOfWeek of daysOfWeek) {
  const timesOnDay = row.times.by_day[dayOfWeek];

  for (let i = 0; i < timesOnDay.length; i++) {
    const startTime = parseFloat(timesOnDay[i][0]);
    if (direction === "After" && startTime < filterTime) {
      return false;
    } else if (direction === "Before" && startTime > filterTime) {
      return false;
    }
  }
}

return true;
```

Note that in the loop, we now return false if anything fails to match the filter rather than returning true if it matches the filter, since we should hide a class if *any* day has a time that doesn't fit our filter

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:21

Thanks for trying this out! This is definitely a useful feature for students trying to find classes that they don't have to wake up too early for, but currently runs into issues.

I've added some notes on how you might be able to fix this, if you'd like.

FilterDatesProblems

```
Uncaught TypeError: Cannot read property 'length' of undefined
    at eval (initializetable.js?7123:597)
    at aggregateFilter (filterui.js?1d53:952)
    at updateDisplayedRows (tabletools.js?8b6b:694)
    at Table.setFilter (tabletools.js?8b6b:157)
    at updateTableWithFilters (filterui.js?1d53:964)
    at FilterManager.activateFilter (filterui.js?1d53:1008)
    at tryApplyingCurrentFilter (filterui.js?1d53:825)
    at HTMLButtonElement.eval (filterui.js?1d53:681)
    at HTMLButtonElement.dispatch (jquery.js:3058)
    at HTMLButtonElement.eventHandle (jquery.js:2676)
```

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:32

The lines after `const buttonText` and before `tryApplyingCurrentFilter` could be simplified and replaced with this:

```
  if (!('day_of_week' in currentFilter)) {
    currentFilter.day_of_week = [];
  }

  const dayOfWeekAlreadyClicked = currentFilter.day_of_week.includes(buttonText);
  if (!dayOfWeekAlreadyClicked) {
    currentFilter.day_of_week.push(buttonText);
  }
```

Less code that's more readable tends to work better!

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:35

`findButtonByText` no longer describes this very well. Perhaps: `findButtonsWithLabels`, and renaming `text` to `labels` would make it clearer.

This used to find one button with the exact text. Now it finds a set of buttons that have text that might match an array of text.

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:36

changed title from **Add multi-day time filters** to **Add multi-day time filters{+ !4+}**

pxpeterxu:
In GitLab by @hsource on Sep 14, 2019, 02:36

changed the description

hsheth2:
Closing for now -- when this is updated, we can reopen the PR.

PR #76, opened 2019-11-20
Updated install scripts after testing for Ubuntu 18.04

hsource:
## Motivation

I wanted to ensure that the install and build scripts are fully working and to document the instructions.

## Fix

The build scripts are imported from one of our other projects, and I ran them manually when setting up this new server.

I added documentation to the [CourseTable development guide](https://docs.google.com/document/d/1M0Gp8Qtaik8roGYYknDDEzAAOwP3YBj0mX1pvCy-uOI/edit#heading=h.wlmwk4swsj5d) as well.


sysadmin/phpmyadmin/config.inc.php
```diff
@@ -0,0 +1,157 @@
+<?php
```
hsheth2: I think this file is included as `sysadmin/phpmyadmin/config.inc.php` and `phpmyadmin/config.inc.php`.

sysadmin/phpmyadmin/config.inc.php
```diff
@@ -0,0 +1,157 @@
...
+ * or at <https://docs.phpmyadmin.net/>.
+ *
+ * @package PhpMyAdmin
+ */
+
+/**
+ * This is needed for cookie based authentication to encrypt password in
+ * cookie
+ */
+$cfg['blowfish_secret'] = 'ThisISPhpMyADMINforAlltheflightdeals\'SNOTREALLY'; /* YOU MUST FILL IN THIS FOR COOKIE AUTH! */
```
hsheth2: Not sure, but this looks like it should be read in via an environment variable.

sysadmin/phpmyadmin/config.inc.php
hsource: Deleted!

sysadmin/phpmyadmin/config.inc.php
hsource: I added a comment
```php7
/**
 * This is needed for cookie based authentication to encrypt password in
 * cookie.
 * 
 * Note that it's OK for this to be embedded in this doc in clear-text:
 * it's used to encrypt cookies at rest. Using the secret requires someone to
 * have also accessed the database and a user's cookies. Per phpMyAdmin docs,
 * it's fine to just fill this in in the file rather than using an environment
 * variable:
 * https://docs.phpmyadmin.net/en/latest/setup.html#manually-creating-the-file
 */
```

PR #84, opened 2019-12-03
Fix bold in AboutModal

ErikBoesen:
We used `<strong>` instead of `</strong>`


hsheth2:
The .ejs file isn't actually in use here -- the .tpl file is the one we want. Integrating these changes into #79.

PR #86, opened 2019-12-03
Hotfix: Facebook credentials in prod

hsheth2:
This patch has already been applied to the server.


hsheth2:
Not exposed, just not loading properly.

Issue #78, opened 2019-12-02
Worksheet-related buttons are still shown when switching seasons


ErikBoesen:
This allows, for example, opening a timetable of all courses in a semester.
<img src="https://user-images.githubusercontent.com/5923662/69997084-5f962180-1521-11ea-889d-9b9a7b3fc0fe.png">

hsheth2:
Wontfix since this isn't an issue in the new site.

PR #87, opened 2019-12-10
Incorporate iCal generation into Coursetable

### LINKED ISSUE #12: Export course worksheet to ICS
In GitLab by @hsheth2 on Sep 5, 2019, 20:49

This could be used to add a calendar subscription which directly adds course times, dates, etc. to a calendar (e.g. Google Calendar) via a live ICS feed.

muavli:
Resolves #12


web/js/worksheets.js
```diff
@@ -128,6 +128,20 @@ export default function WorksheetManager() {
         $('body').append($iframe);
       });
 
+    $('.download-ics-btn')
+      .off('click')
+      .on('click', () => {
+        ociIds = worksheetOciIds.join(',');
+        const $iframe = $(
+          '<iframe istyle="width: 0; height: 0" id="icslink"></iframe>'
```
hsheth2: Can we just use an `<a href>` instead of an `<iframe>`?

web/js/init.js
```diff
@@ -460,6 +460,13 @@ $(document).ready(() => {
     placement: 'bottom',
     container: 'body',
   });
+
+  $('.download-ics-btn').tooltip({
+    title: 'Download ICS File',
```
hsheth2: People probably don't know what an ICS file is, so it might be better to say "Download as Calendar" or something.

web/templates/BluebookPerUser.tpl
```diff
@@ -242,6 +242,7 @@ for compressibility purposes by tools/build.php
 					<span class="btn-group">
 						<button class="btn list-table-btn hide"><i class="icon-th"></i></button>
 						<button class="btn download-csv-btn hide"><i class="icon-download-alt"></i></button>
+            <button class="btn download-ics-btn hide"><i class="icon-calendar"></i></button>
```
hsheth2: Given that the ICS generation only works for the 202001 season, would it possible to hide/disable the button when on other seasons?

muavli: I'm thinking of doing this in the worksheets.js file. Is there an advantage to doing it here?

hsheth2: You're right, it's probably better to do it there.

muavli: So, I did a little experimentation. If we toggle the ics button in BluebookPerUser.tpl, we need to refresh the page in order to see only two buttons. But if in worksheets.js, it's more dynamic but when we see only two buttons, the right edges of the csv download button aren't rounded. Combining the two results in the same effect as having the season check in BluebookPerUser.tpl. This assumes, of course, I correctly implemented the above. What are your thoughts?

muavli:
Also, apologize for failing prettier again. My VSCode does not seem to like leaving a space after function and before an opening parenthesis.

muavli:
Noticed something pretty interesting. If we navigate to different seasons via url, then the ics button appears and disappears correctly (with the right edge of csv button still not being rounded). But if we navigate to different season by the dropdown menu, the ics button doesn't reappear if we switch back to spring 2020. Not sure how we can work around this....

web/templates/BluebookPerUser.tpl
hsheth2: This approach seems better than putting it in BluebookPerUser.tpl. 

I don't think the rounding is a big issue. It might be fixed by putting the calendar button in the middle of the three, but honestly not worth fixing.

hsheth2:
> Noticed something pretty interesting. If we navigate to different seasons via url, then the ics button appears and disappears correctly (with the right edge of csv button still not being rounded). But if we navigate to different season by the dropdown menu, the ics button doesn't reappear if we switch back to spring 2020. Not sure how we can work around this....

This is probably related to #78, so we can just ignore it for this PR.

PR #101, opened 2020-02-08
Remove lines between rows

ErikBoesen:


ErikBoesen:
Sorry this is super late—I think we decided against that because it distinguishes which part of the page is "folded" or not.

hsheth2:
Sounds good -- I'll go ahead and merge it.

PR #105, opened 2020-02-29
Update init.js to fix worksheet buttons bug

maxyuan6717:
A working fix for worksheet buttons that shouldnt be there. Just removes extraneous buttons and keeps all courses view.

Still trying to fix intended behavior by keeping worksheet view open when changing seasons


hsheth2:
Closing since we probably won't fix this

PR #108, opened 2020-07-26
Basic FB integration

jlv34:


jlv34:
Still need to update UI to reflect whether user is logged into FB

PR #116, opened 2020-10-01
Umami analytics

### LINKED ISSUE #112: Metrics via umami
Setup [umami](https://umami.is/) and integrate with the frontend - may want to use something like this https://stackoverflow.com/a/58205291/5004662

hsheth2:
Closes #112.


hsheth2:
fyi made a couple edits for the hosted umami instance - going to merge it in

Issue #95, opened 2020-01-18
Disconnect Facebook does not work completely


hsheth2:
Via a user report:
> disconnecting and reconnecting didn't remove people I'm no longer Facebook friends with

muavli:
@hsheth2 this should be closed by #190 right?

hsheth2:
yup

PR #118, opened 2020-10-03
Facebook integration bug fixes

### LINKED ISSUE #111: Facebook integration bug
Fix the bug with facebook integration - should not call FetchFacebookData every time, but should support a refresh button; maybe we should automatically do it once per day or something?

hsheth2:
Closes #111


hsheth2:
Still TODO are #95 (disconnect) and refresh option

PR #123, opened 2020-10-16
Worksheet updates

maxyuan6717:
-Add separate login page for worksheet
-Update index.php to support redirect after login success
-Add login button to catalog page
-Redirect to the page u logged in from for both catalog and worksheet


hsheth2:
LGTM

PR #133, opened 2020-10-18
About page

hsheth2:


hsheth2:
@nsnave this isn't necessary but it might be cleaner to do something similar to this: https://github.com/coursetable/coursetable/blob/cc7d032cd11f2ef1ae4dc3b70fd7e09e6c21cb25/frontend/src/pages/FAQ.js#L50-L222

maxyuan6717 (approved):
Looks good! I just fixed the prettier formatting to pass the checks. FYI for future PRs, we use prettier to format and standardize our code. If ur using VScode, there's an extension for it!

PR #135, opened 2020-10-18
Calendar updates

maxyuan6717:
-Remove course time from a calendar event
-"Hide" other courses when ur hovering over a specific one by decreasing opacity
-Bring the event to the front on hover
-Slightly enlarge event on hover


hsheth2 (approved):
might be nice to make the animation a bit faster, but other than that it looks good

PR #143, opened 2020-10-19
Fix android worksheet bug

maxyuan6717:
-Remove @media(hover: none)
-Set max-height if not on mobile in a better way


hsheth2 (approved):
Seems good but I can't really test it myself

Once it's deployed we can ask @nsnave to try it

PR #148, opened 2020-10-20
Use react code-splitting

hsheth2:
Not yet sure if this is a good move

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -1,4 +1,10 @@
-import React, { useEffect, useState, useCallback } from 'react';
+import React, {
+  useEffect,
+  useState,
+  useCallback,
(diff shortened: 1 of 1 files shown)


hsheth2:
@maxyuan6717 curious if you think this is a good idea

I don't really know what the tradeoffs are here

maxyuan6717:
Yah I'm not too familiar with the tradeoffs either, but the overall idea of code splitting seems beneficial in our case. I think it's worth a shot

hsheth2:
Alright i deployed it to the beta site to test it - I feel like it doesn't really help that much since our site isn't that big

hsheth2:
Seems like it broke the home page search, so closing for now

PR #154, opened 2020-10-24
Fb dropdown updates

maxyuan6717:
-Fix bug with showing 'x' or 'v' indicator for FB friends dropdown
-Constrain "Friends' courses" placeholder on one line and add ellipsis if too long
-Allow event boxes to scale outside of rbc-time-content container if on the edge


hsheth2 (approved):
LGTM - the odd characters in the css rules are kinda worrying but I'll assume its fine

PR #177, opened 2020-10-30
Remove worksheet, catalog animations; Speed up modal animation

maxyuan6717:


hsheth2 (changes requested):
looks good - can we also deal with the animation when opening the course modal?

I think a good rule of thumb is that 100ms or less animations are perceived as instant

hsheth2 (approved):

PR #180, opened 2020-11-01
Static catalog

kevinhu:
- Adds `/api/static` endpoint for accessing static files (NetID auth required):
   - `/api/static/seasons.json`: list of all available seasons
   - `/api/static/catalogs/{season}.json`: catalog per season
- Fetches and populates static files on API server start (set `OVERWRITE_CATALOG` environment parameter to update each catalog file, otherwise existing catalogs are skipped)
- Adds `/api/catalog/refresh` endpoint to refresh the catalog remotely. Can be called without authentication in development, otherwise set `FERRY_SECRET` in environment and specify `x-ferry-secret` header when calling.


hsheth2 (approved):

hsheth2:
Made some minor changes with the env handling

Also Promise.all -> Promise.allSettled so we wait for all promises to be resolved

PR #183, opened 2020-11-02
Fix worksheet expanded and remove infinite scroller from search results

maxyuan6717:
-Replace WorksheetExpandedList with just SearchResults
-Remove infinitescroller from SearchResults
-Fix support for multiple seasons
TODO: add back season and fb friend switcher to expanded worksheet list view


frontend/src/components/WorksheetExpandedList.js
```diff
@@ -32,148 +34,13 @@ const WorksheetExpandedList = ({
...
-        </div>
-        {/* Render HTML of items */}
-        <div className={styles.results_list_container}>{items}</div>
-      </Container>
-    </div>
+    <SearchResults
+      data={courses}
+      showModal={showModal}
+      expanded={cur_expand === 'list'}
+      isLoggedIn={false}
```
hsheth2, line 42: wat

hsheth2 (approved):

frontend/src/components/WorksheetExpandedList.js
maxyuan6717: oops the isLoggedIn should always be true cuz u can only access the worksheet if logged in. Other than that, it's just required props for SearchResults

PR #185, opened 2020-11-02
Course Modal Updates

maxyuan6717:
-Add number of evals in each course, both, prof column to the multi toggle header
-Add n/a ratings to fill in blanks
-Evals in both column also show up in the course and prof columns


hsheth2 (approved):
looks good

We'll eventually need up update the catalog enrollment data column now this is merged https://github.com/coursetable/ferry/pull/37

PR #190, opened 2020-11-03
Facebook Disconnect and Sync updates

maxyuan6717:


hsheth2:
@muavli lmk when this is ready to merge - it looks like it's in good shape to me

PR #194, opened 2020-11-04
Fix worksheet performance issues

maxyuan6717:
-Convert WorksheetAccordion to functional component
-Add useMemos and useCallbacks everywhere


maxyuan6717:
worksheets should be buttery smooth even with 50+ courses

hsheth2 (approved):

PR #197, opened 2020-11-04
Data provider based on static files

hsheth2:


frontend/src/pages/Search.js
```diff
@@ -162,239 +155,320 @@ function Search({ location, history }) {
...
-        return preprocess_courses(x);
+  // If the search query is still loading
+  if (coursesLoading) {
+  } else if (searchData) {
+    // Scroll down to catalog when search is complete for mobile view
+    // TODO: this is called too frequently
+    if (isMobile) {
+      scroller.scrollTo('catalog', {
+        smooth: true,
+        duration: 500,
```
hsheth2, line 458: @maxyuan6717 any idea what to do here now that we have live-updating results?

maxyuan6717: Maybe adding back the submit button on mobile and only filtering/scrolling on submit? So like only update searchConfig on submit

hsheth2: It feels like it would complicate data flow a bit - what if we scrolled on first render + anytime they hit "enter" in the search form to submit it, but kept the live updating?

maxyuan6717: ohhh I like that

hsheth2: kk will do that

maxyuan6717 (approved):

PR #206, opened 2020-11-06
darkmode

nsnave:
Used darkreader library for easy conversion of site (https://github.com/darkreader/darkreader)


hsheth2:
dang that's sick

frontend/src/components/Navbar.js
```diff
@@ -46,6 +53,32 @@ function CourseTableNavbar({ isLoggedIn }) {
     window.location.pathname = '/';
   };
 
+  // DarkMode or LightMode
+  setFetchMethod(window.fetch);
+
+  // Checks user preferences for dark mode
+  if (!window.localStorage.getItem('darkmode')) {
+    var temp =
```
hsheth2, line 61: Generally with JS we want to use `let` or `const` - there's a bunch of weird scoping issues with `var`

frontend/src/components/Navbar.js
```diff
@@ -46,6 +53,32 @@ function CourseTableNavbar({ isLoggedIn }) {
     window.location.pathname = '/';
   };
 
+  // DarkMode or LightMode
+  setFetchMethod(window.fetch);
```
hsheth2, line 57: what does this do?

frontend/src/components/DarkModeButton.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * DarkMode Button
+ * @prop darkModeEnabled - boolean to determine which image to display (true for "sun"; false for "moon")
+ */
+
+function DarkModeButton({ darkModeEnabled = false }) {
+  return (
+    <span>
+      <img
+        src={darkModeEnabled ? sun : moon}
```
hsheth2, line 17: We already have the react-icons library and we're mostly using icons from font-awesome.

The FaSun and FaMoon/FaRegMoon might be a good fit, and then we can avoid the extra png's

maxyuan6717:
Thoughts on sticking the toggle under the me dropdown?

hsheth2:
I kinda like it in the navbar - it's much more discoverable there. It also reminds me of the interface in umami, which I think is pretty clean.

frontend/src/components/Navbar.js
nsnave: not 100% sure on all the details but sometimes the darkreader library needs to access "a cross-origin resource" and without this line it throws an error

hsheth2 (approved):

Issue #201, opened 2020-11-04
Move search options to above catalog


hsheth2:
Google Flights has a really clean UI in this regard

PR #213, opened 2020-11-12
Search box style changes

hsheth2:
Someone complained that the search box is really hard to see in dark mode - this is a quick fix for it, but we can also just reject and wait for #201 first


maxyuan6717 (approved):
Sure we can add this. Rn, I'll keep the current styles and just get the overall layout figured out first.

PR #216, opened 2020-11-12
Generate grid html on the fly

hsheth2:
Avoiding pre-computing it should make it faster, especially when viewing a list

Would probably be even better to split grid and list into two different components tho


maxyuan6717 (approved):
Yah after I get this new search UI set up, we can refactor Search.js and SearchResults.js

PR #218, opened 2020-11-13
Sentry.io setup

hsheth2:


avgupta456 (approved):
Looks good to me. I checked and we did it the exact same way at Gatherly.

PR #221, opened 2020-11-13
Catalog improvements

maxyuan6717:
-Move course info popover to left when catalog not expanded
-Always show all catalog columns
-Add more detailed tooltips to ratings and enrollment columns
-Change order of columns
-Add number of search results


frontend/src/components/SearchResultsItem.js
```diff
@@ -294,63 +266,87 @@ const SearchResultsItem = ({
...
+          ))}
+          {course.areas.map((area) => (
+            <Badge
+              variant="secondary"
+              className={Styles.tag + ' my-auto'}
+              key={key++}
+              style={{
+                color: skillsAreasColors[area],
+                backgroundColor: chroma(skillsAreasColors[area])
+                  .alpha(0.16)
```
hsheth2, line 328: A couple people mentioned that we should make these colors have even more contrast so that they really pop.

hsheth2 (commented):
Some quick initial comments - will test this out on my laptop in a bit.

frontend/src/components/SearchResultsItem.js
maxyuan6717: yah changing from bezier to scale makes the colors look just like old table

frontend/src/components/SearchResultsItem.js
```diff
@@ -283,74 +253,98 @@ const SearchResultsItem = ({
...
+              ? workloadColormap(course.average_workload).darken(3).css()
               : '#b5b5b5',
             backgroundColor: course.average_workload
-              ? chroma(workloadColormap(course.average_workload))
+              ? workloadColormap(course.average_workload)
               : '#ebebeb',
           }}
           className={Styles.rating_cell + ' m-auto'}
         >
           {course.average_workload ? course.average_workload.toFixed(1) : 'N/A'}
```
hsheth2, line 264: If all three are N/A, the data looks a bit packed - can we make the N/A text non-bolded and/or a smaller font size

image

hsheth2 (approved):
looks good!

This should knock off a whole bunch of issues which I'm pretty excited about

PR #222, opened 2020-11-14
Course info popover in worksheet calendar

maxyuan6717:


hsheth2 (approved):
Looks good

The popup show delay of 500ms feels a bit slow to me, but I leave the final decision to you

maxyuan6717:
Yah I'll bump it down to 300ms, and we can always adjust if ppl think it gets in the way

PR #225, opened 2020-11-14
preserve user settings in system storage

nsnave:


frontend/src/pages/Search.js
```diff
@@ -81,9 +83,13 @@ function Search({ location, history }) {
...
   const searchTextInput = useRef(null);
-  const [searchText, setSearchText] = useState('');
+  setSSObjectIfEmpty('searchText', '');
+  const [searchText, setSearchText] = useState(getSSObject('searchText'));
   // Is the search form  collapsed?
-  const [collapsed_form, setCollapsedForm] = useState(false);
+  setSSObjectIfEmpty('collapsed_form', false);
+  const [collapsed_form, setCollapsedForm] = useState(
+    getSSObject('collapsed_form')
+  );
```
hsheth2, line 92: could make `getSSObject` take a second argument with the default value - that way we can combine `setSSObjectIfEmpty` with `getSSObject`

nsnave: good idea

frontend/src/pages/Search.js
```diff
@@ -124,24 +132,56 @@ function Search() {
...
-  const [select_seasons, setSelectSeasons] = useState([
-    { value: '202101', label: 'Spring 2021' },
-  ]);
-  const [select_skillsareas, setSelectSkillsAreas] = useState();
-  const [select_credits, setSelectCredits] = useState();
-  const [select_schools, setSelectSchools] = useState([]);
-  const [select_subjects, setSelectSubjects] = useState([]);
+  const [select_sortby, setSelectSortby] = useState(
+    getSSObject('select_sortby')
+  );
```
hsheth2, line 154: This entire thing feels really repetitive and that worries me

What if we added a `useSessionStorageState` React hook that takes a name and default value and manages all of this stuff internally
```JavaScript
function useSessionStorageState(name, default) {
  setSSObject(name, default, true);
  const [value, setValue] = useState(getSSObject(name));
  useEffect(() => {
    setSSObject(name)
  }, [value]);
  return [value, setValue];
}
```

hsheth2 (changes requested):

hsheth2 (approved):

PR #261, opened 2020-11-16
Fix react warnings

maxyuan6717:
this also improves worksheet performance a bit cuz it cuts down on the calendar rerenders

### DIFF
--- a/frontend/src/components/CourseModalOverview.js
+++ b/frontend/src/components/CourseModalOverview.js
@@ -106,18 +106,23 @@ const CourseModalOverview = ({ setFilter, filter, setSeason, listing }) => {
     },
   });
   // Hold list of evaluation dictionaries
-  let evaluations = [];
+  let evaluations = useMemo(() => {
+    return [];
+  }, []);
   // Hold HTML code that displays the list of evaluations
 
   // Holds Prof information for popover
-  let prof_info = {};
-  listing.professor_names.forEach((prof) => {
-    prof_info[prof] = {
-      num_courses: 0,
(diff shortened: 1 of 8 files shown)


frontend/src/components/WeekSchedule.js
```diff
@@ -132,10 +132,12 @@ function WeekSchedule({ showModal, courses, hover_course }) {
...
-  const event_style = {
-    width: '100%',
-    height: '100%',
-  };
+  const event_style = useMemo(() => {
+    return {
+      width: '100%',
+      height: '100%',
+    };
+  }, []);
```
hsheth2, line 140: Move this outside the component function?

frontend/src/components/CourseModalOverview.js
```diff
@@ -106,18 +106,23 @@ const CourseModalOverview = ({ setFilter, filter, setSeason, listing }) => {
     },
   });
   // Hold list of evaluation dictionaries
-  let evaluations = [];
+  let evaluations = useMemo(() => {
+    return [];
+  }, []);
```
hsheth2, line 111: this seems kinda sus - all the logic to compute the evaluations should be inside here right

hsheth2 (changes requested):

frontend/src/components/CourseModalOverview.js
maxyuan6717: oh yah im tweaking. just moved `let evaluations = []` into the useMemo for items

hsheth2 (approved):
LGTM

PR #262, opened 2020-11-16
Worksheet/Calendar performance improvements

maxyuan6717:
-Scroll performance always smooth
-Hide/Show courses much more efficient. Should feel instantaneous
-Worksheet List Item has its own component
-Disable hover interaction if >30 courses in worksheet (up until 30, it is seamless on my machine)

### DIFF
--- a/frontend/src/components/WeekSchedule.js
+++ b/frontend/src/components/WeekSchedule.js
@@ -179,7 +179,7 @@ function WeekSchedule({ showModal, courses, hover_course }) {
     (event) => {
       const border = '1)';
       let style;
-      if (hover_course && hover_course.crn === event.listing.crn) {
+      if (hover_course && hover_course === event.listing.crn) {
         style = {
           backgroundColor: event.listing.color.concat('1)'),
           borderColor: event.listing.color.concat(border),
@@ -225,6 +225,24 @@ function WeekSchedule({ showModal, courses, hover_course }) {
       : moment().hour(18).minute(0).toDate();
   }, [ret_values]);
 
+  const customEventMemo = useMemo(() => {
+    return { event: customEvent };
+  }, [customEvent]);
+
+  const onSelectEventCallback = useCallback(
+    (event) => {
+      return showModal(event.listing);
+    },
+    [showModal]
+  );
+
+  const eventPropGetterCallback = useCallback(
+    (event) => {
+      return eventStyleGetter(event);
+    },
+    [eventStyleGetter]
+  );
+
   return (
     <StyledCalendar
       // Show Mon-Fri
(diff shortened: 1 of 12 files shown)


frontend/src/components/WeekSchedule.js
```diff
@@ -78,8 +79,6 @@ function WeekSchedule({ showModal, courses, hover_course }) {
     let id = 0;
```
hsheth2, line 79: getting rid of this variable would help keep stable keys, which might also help performance

frontend/src/pages/Worksheet.js
```diff
@@ -101,36 +101,15 @@ function Worksheet() {
...
-      else {
-        let temp = [...hidden_courses];
-        temp.splice(isHidden(season_code, crn), 1);
-        setHiddenCourses(temp);
-      }
+    (crn) => {
+      let new_hidden_courses = Object.assign({}, hidden_courses);
+      if (hidden_courses[crn]) new_hidden_courses[crn] = false;
+      else new_hidden_courses[crn] = true;
+      setHiddenCourses(new_hidden_courses);
```
hsheth2, line 110: you can actually pass a function into setHiddenCourses and remove the direct dependency on hidden_courses

frontend/src/components/WorksheetListItem.module.css
```diff
@@ -0,0 +1,25 @@
+.list_text {
+  overflow: hidden;
+  transition: 0.3s;
```
hsheth2, line 3: we're probably transitioning on too many properties here

frontend/src/components/WorksheetListItem.module.css
```diff
@@ -0,0 +1,25 @@
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.bookmark {
+  position: absolute;
+  top: -16px;
+  right: 0%;
+  z-index: 1;
+  transition: 0.3s;
```
hsheth2, line 19: ditto

hsheth2 (changes requested):
Ideally we'd get the hover interaction to work smoothly even with >30 courses - in fact, it's the most useful when you have a lot of classes

frontend/src/pages/Worksheet.js
maxyuan6717: not too sure what u meant here, so I just pushed what my understanding was

frontend/src/pages/Worksheet.js
```diff
@@ -101,17 +101,24 @@ function Worksheet() {
...
-      setHiddenCourses(new_hidden_courses);
+      return new_hidden_courses;
     },
     [hidden_courses]
   );
 
+  // Hide/Show this course
+  const toggleCourse = useCallback(
+    (crn) => {
+      setHiddenCourses(updateHiddenCourses(crn));
```
hsheth2: Ah not quite what I meant originally - you can actually pass a function like this:
```javascript
setHiddenCourses((old_hidden_courses) => {
    // do transformation on old_hidden_courses
    return new_hidden_courses;
})
```

I just made the change myself because it's pretty quick.

hsheth2 (approved):

PR #271, opened 2020-11-17
Fix Hide Course

maxyuan6717:
Hiding course completely removes it from calendar, so other courses expand to take up missing space


hsheth2 (approved):
LGTM

PR #272, opened 2020-11-17
Darkmode compatible navbar toggle

maxyuan6717:


frontend/src/components/Navbar.js
```diff
@@ -45,6 +45,18 @@ const StyledNavLink = styled(NavLink)`
...
 
+const StyledNavToggle = styled(Navbar.Toggle)`
+  border-color: ${({ theme }) => theme.border} !important;
+  .navbar-toggler-icon {
+    background-image: ${({ theme }) =>
+      "url(\"data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' width='30' height='30' viewBox='0 0 30 30'%3e%3cpath stroke='" +
+      (theme.theme === 'light'
+        ? 'rgba(69, 69, 69, 1)'
+        : 'rgba(219, 219, 219, 1)') +
+      "' stroke-linecap='round' stroke-miterlimit='10' stroke-width='2' d='M4 7h22M4 15h22M4 23h22'/%3e%3c/svg%3e\")"};
```
hsheth2, line 56: This seems really ugly - is there no way to set the fill color or filter via css?

maxyuan6717: not unless u replace the icon with one from like font-awesome or something. Yah idk y bootstrap does this

hsheth2: ah fair enough

hsheth2 (approved):

PR #291, opened 2020-11-19
Add link icons to syllabus and locations in modal

maxyuan6717:


frontend/src/components/CourseModalOverview.js
```diff
@@ -550,8 +553,10 @@ const CourseModalOverview = ({ setFilter, filter, setSeason, listing }) => {
                   target="_blank"
                   rel="noopener noreferrer"
                   href={listing.syllabus_url}
+                  className="d-flex"
                 >
                   {listing['course_code']}
```
hsheth2, line 558: Maybe we should change the text to "view syllabus" or something?

Also we should move the entire syllabus row up in the modal - probably should be the first item right?

hsheth2 (approved):
couple suggestions

no need to re-request review - this should be pretty simple

PR #292, opened 2020-11-20
Remove worksheet's inner scroll

maxyuan6717:


hsheth2 (changes requested):
Looks good except for the artifact that shows up above the option picker when scrolled down in light mode

image

hsheth2 (approved):

PR #297, opened 2020-11-22
Add sort options to expanded worksheet

maxyuan6717:


frontend/src/components/WorksheetExpandedList.js
```diff
@@ -38,13 +55,56 @@ const WorksheetExpandedList = ({
...
+      courses,
+      [
+        (listing) => !!listing[key],
+        (listing) => listing[key],
+        (listing) => listing.course_code,
+      ],
+      ['desc', order_asc ? 'asc' : 'desc', 'asc']
+    );
+    return listings;
+  }, [courses, select_sortby, sort_order]);
```
hsheth2, line 99: Is there a way to refactor this logic out - don't want to duplicate logic between the search and here

hsheth2 (approved):

PR #315, opened 2020-11-26
Fix search form bugs

maxyuan6717:
-Don't preserve visibility of search form
-Fix bug where the "show search form" icon isn't clickable on displays where the form is taller than the screen height
-Use overall same professor ratings in overall rating slider
-Visually update rating sliders when resetting filters

Issue was the positioning of the form being weird when the screen is too short, causing the icon to be placed behind something.

### DIFF
--- a/frontend/src/pages/Search.js
+++ b/frontend/src/pages/Search.js
@@ -94,11 +94,8 @@ function Search() {
   const [ordering, setOrdering] = useSessionStorageState('ordering', {
     course_code: 'asc',
   });
-  // State to reset sortby select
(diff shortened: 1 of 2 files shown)


nsnave (approved):
Look goods, another bug I just realized is in the workload/rating slider. When one moves the slider and then clicks reset, the slider doesn't get updated visually.

hsheth2 (approved):

maxyuan6717:
> Look goods, another bug I just realized is in the workload/rating slider. When one moves the slider and then clicks reset, the slider doesn't get updated visually.

PR #316, opened 2020-11-26
Reduce data loading

### LINKED ISSUE #227: Avoid loading all seasons in worksheet until necessary
We don't need to load like 6 seasons on initial load

hsheth2:
Closes #227
- [x] Improve catalog performance by not loading static files for every season in worksheet, but instead only for the relevant seasons
- [x] Load seasons one at a time in the worksheet


maxyuan6717 (approved):
thanks for cleaning this up! only thing is that when u remove the last course of a season (but there r still other courses in other seasons), the course doesn't get removed visually; it does get removed on refresh tho.

PR #320, opened 2020-11-29
Make eval filter multitoggle focusable

maxyuan6717:
Toggle between eval filters using arrow keys on focus. Should I add some UI indication that the user has focused on the multitoggle?


hsheth2 (commented):
Code seems fine

I should've been more clear - I think the linked issue was referring to this part
image

PR #383, opened 2020-12-19
New auth part 1

hsheth2:
Adds CAS auth support to the API, but it doesn't do anything with the auth yet. This way we can test to make sure this part works on the live site first.

Afterwards we can replace the current PHP auth with this system (not built yet). Note that this switch will invalidate all active sessions and log everyone out, so we'll want to do it after the initial shopping period ends. 

Note that I also forked the passport-cas library: https://github.com/coursetable/passport-cas (docs are not updated).


kevinhu (approved):
Wow, using passport-cas makes auth surprisingly clean. Can't wait for the day we won't have to check Duo every time we open up the site!

PR #385, opened 2020-12-19
Better empty worksheet state and always show all fb friends

maxyuan6717:
TODO for both worksheet list and mobile
-Improve empty worksheet state
-Always show all FB friends even if they don't have courses for this season


hsheth2 (approved):
Code looks good. I haven’t tested it but trust that you did :)

At one point we had a bug in the mobile where selecting a friend with no courses made it impossible to select anyone else, but not sure if that was due to the localstorage or an actual bug. In any case, can you check that we don’t have that bug anymore and then merge this in?

maxyuan6717:
yup, I've done some testing with a secondary account, and that bug seems to be fixed now!

PR #394, opened 2020-12-22
Add hide/show all and expand worksheet btns

maxyuan6717:
don't know if this is the look we're going for


hsheth2 (approved):
Code LGTM

@mich-m-li  / @itsjoshthedeveloper this is what it looks like - let us know what you think: 
image

itsjoshthedeveloper:
Is the expand button to go to list view?

maxyuan6717:
yah

itsjoshthedeveloper:
I think it would be more intuitive to have it say "go to list view" rather than "expand". I also think it would be more intuitive to put it above the card and keep it text because it doesn't have anything to do with changing the list of courses in your worksheet but instead changing what view the worksheet is in overall.

itsjoshthedeveloper:
I moved the expand button to the top and changed the text to "Go to list view"

Default state
image

Hover state
image

@maxyuan6717 does that seem fine?

itsjoshthedeveloper:
also, is there a reason why it seems that the conflicting files aren't pulling the latest commit for those files? for ex, in `Worksheet.js` it doesn't have the lines importing `posthog` and `FaCompressAlt`, `FaExpandAlt`...

hsheth2:
Fixed the conflicts - the imports were reordered in the master branch which caused the conflict

mich-m-li:
For the "Hide All" button, is it reversible? My assumption is that button will hide all courses from the Calendar. 

Ideally, I think we should allow users to be able to go back and forth between seeing everything and nothing--in addition to the worksheet saving the user's combination of hidden and shown courses.

itsjoshthedeveloper:
> For the "Hide All" button, is it reversible? My assumption is that button will hide all courses from the Calendar. 
> 
> 
> 
> Ideally, I think we should allow users to be able to go back and forth between seeing everything and nothing--in addition to the worksheet saving the user's combination of hidden and shown courses.

yup max made it reversible! @hsheth2 is this good to merge?

maxyuan6717:
sry for the holdup, my docker is being sus again and not building the frontend properly. It looks good from the screenshots, but could you send a pic of the full screen so I can see what it looks like in context with the calendar? thanks!

itsjoshthedeveloper:
@maxyuan6717 @mich-m-li @hsheth2 

https://user-images.githubusercontent.com/34778271/103595157-c4604000-4ec8-11eb-82bd-ad05cb29dbfc.mp4

maxyuan6717:
3 things:
1. dunno how everyone else feels about this, but the list of courses starting below where the calendar starts looks a bit wack to me
2. I feel like the hide button isn't that important of a feature compared to season and FB, so it might be overemphasized by taking up 100% width
3. dunno if you've taken into this account, but I would test the scrolling with a bunch of courses in ur worksheet to make sure they don't overlap the "go to list view" link. My styling here is kinda weird.

My code still won't run so I gotta hit up @hsheth2 to get this sorted out lul

itsjoshthedeveloper:
> 3 things:
> 
> 1. dunno how everyone else feels about this, but the list of courses starting below where the calendar starts looks a bit wack to me
> 2. I feel like the hide button isn't that important of a feature compared to season and FB, so it might be overemphasized by taking up 100% width
> 3. dunno if you've taken into this account, but I would test the scrolling with a bunch of courses in ur worksheet to make sure they don't overlap the "go to list view" link. My styling here is kinda weird.
> 
> My code still won't run so I gotta hit up @hsheth2 to get this sorted out lul

1. I can try putting it inside the SurfaceComponent?
2. since it's the same height, imo it doesn't feel like it overshadows the season and fb friends
3. making it so that the link is inside the sticky div as well

itsjoshthedeveloper:
@maxyuan6717 @mich-m-li @hsheth2
my b i keep committing before asking yall oops

https://user-images.githubusercontent.com/34778271/103611451-50d22900-4ef0-11eb-94be-112413b9955d.mp4

hsheth2:
My 2c:
- Still feels like the show/hide is overemphasized, but I don't have a better solution
- The arrow icon in the link feels unintuitive - it's not really left/right anymore

I'll defer to @maxyuan6717 for actual code review / decision making since he basically owns frontend

maxyuan6717:
Yah this looks pretty good! Only thing is if we still want the expand icons for the calendar cuz now it doesn't match the expand button for the list on the right

itsjoshthedeveloper:
i'd say it makes sense as it is since i see the `none` view as `calendar` view and the expand button just makes it larger

mich-m-li:
A note on how courses look when "hidden": the gray out and eye strikethrough icon make them look, "canceled"? We could copy Figma's system in which only hidden courses are denoted with an icon, and shown courses display the eye icon upon hover:
image none hidden
image on hover 
image hidden

itsjoshthedeveloper:
@mich-m-li i think this is a great idea, but might not be this PR might not be the best place for it. i can start a new issue and branch for it

PR #414, opened 2020-12-30
Deepsource bug fixes

hsheth2:


maxyuan6717 (approved):
LGTM do we need to do anything about the check failures

hsheth2:
Those are all just warnings - eventually we should fix them but it’s fine for now

PR #419, opened 2021-01-04
Auth part 2

nsnave:


frontend/src/components/Navbar.tsx
```diff
@@ -207,8 +207,7 @@ function CourseTableNavbar({
                       onClick={() => {
                         posthog.capture('login');
 
-                        window.location.href =
-                          '/legacy_api/index.php?forcelogin=1';
+                        window.location.href = '/api/auth/cas';
```
hsheth2, line 210: Would be good to always specify the redirect URL instead of relying on the default (here and in other tsx files)

proxy/proxy.js
```diff
@@ -49,18 +49,28 @@ const authSoft = (req, _, next) => {
...
         return res.status(403).send('request missing authentication');
       }
 
-      if (!data.evaluationsEnabled) {
+      let evals_enabled = await axios
+        .get(`${php_uri}/CheckEvals.php`, {
+          headers: {
+            'x-coursetable-netid': data.id,
+          },
+        })
```
hsheth2, line 68: This is a bit ugly but a good solution for now

Can you add a TODO here - eventually we'll want to add evals information to API natively and avoid the second request within proxy.

proxy/proxy.js
```diff
@@ -49,18 +49,28 @@ const authSoft = (req, _, next) => {
...
-    .then(({ data }) => {
-      if (!data.success) {
+    .then(async ({ data }) => {
+      if (!data.auth) {
         // Return 403 forbidden if the request lacks auth.
         return res.status(403).send('request missing authentication');
       }
 
-      if (!data.evaluationsEnabled) {
+      let evals_enabled = await axios
```
hsheth2, line 63: `const` instead of `let`

proxy/proxy.js
```diff
@@ -74,6 +84,7 @@ const authHard = (req, res, next) => {
 
 // Setup all the proxy routes.
 
+app.use('/legacy_api', authSoft);
```
hsheth2, line 87: for some reason github doesn't let me add comments anywhere in the file

can you add another TODO - eventually we'll want to remove `authSoft` for requests to `/api`

web/includes/ProjectCommon.php
```diff
@@ -81,23 +76,12 @@ public static function createYalePlusMysqli()
...
+    public static function getID()
     {
-        if ($force) {
-            phpCAS::forceAuthentication();
-        }
-
-        $netId = null;
-        if (phpCAS::isAuthenticated()) {
-            $netId = phpCAS::getUser();
+        $netId = $_SERVER['HTTP_X_COURSETABLE_NETID'];
```
hsheth2, line 81: nice!

web/GenerateIcs.php
```diff
@@ -203,7 +203,7 @@ function createIcsFile($ociIds, $ociIdData, $season, $outputStream)
     $ociId = (int) $ociId;
 }
 
-$netId = ProjectCommon::casAuthenticate(true);
+$netId = ProjectCommon::getID();
```
hsheth2, line 206: `casAuthenticate(true)` previously would redirect if the user wasn't logged in, whereas `casAuthenticate(false)` would return null when the user wasn't logged in. I don't think this will cause any issues, but you probably know better than I do.

hsheth2 (approved):
A couple minor comments but overall looks good

web/GenerateIcs.php
nsnave: yeah, you're right and I don't think any of the calls casAuthenticate(true) actually get called so it shouldn't be a problem, Only Table.php and GenerateIcs.php ever call it, which I don't think are used (?), but even if they are used I don't think it would be an issue.

proxy/proxy.js
nsnave: Why will we remove authSoft?

hsheth2: Right now the API still looks at the header info for auth information. The flow is currently this: incoming request -> request /api/auth/check -> set headers -> proxy request along to /api/whatever

But the API can just inspect the cookie and figure out auth info - it doesn't really need the x-coursetable-* headers anymore. Once we make that change, we won't need to call authSoft for /api

web/GenerateIcs.php
hsheth2: sounds good, thanks for looking into it

proxy/proxy.js
nsnave: Makes sense, that's really interesting

kevinhu (approved):

hsheth2 (approved):

PR #426, opened 2021-01-05
Fix linting

hsheth2:
There was another part of this that I accidentally pushed directly to master

### DIFF
--- a/api/.prettierrc.yaml
+++ b/api/.prettierrc.yaml
@@ -1,2 +1,1 @@
 singleQuote: true
-trailingComma: es5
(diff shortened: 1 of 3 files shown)


itsjoshthedeveloper:
should i add in the `singleQuote: true` in the `.eslintrc.js` on this branch too or will it conflict with the same change on the `hide-all` branch?

hsheth2:
These changes should alleviate the need for modifying `.eslintrc.js`. If not, we can just bundle that change when `hide-all` gets merged

itsjoshthedeveloper (approved):

PR #445, opened 2021-01-09
Hide icon only on hover & clean up hide all button styling

itsjoshthedeveloper:
@mich-m-li


itsjoshthedeveloper:
hide all button before:
hide-all-before

hide all button after:
hide-all-after

itsjoshthedeveloper:
hide icon on hover has some lag... @maxyuan6717 know how to fix this?

hide-icon-lag

maxyuan6717:
the worksheet stuff is often plagued by a bunch of rerenders so that might be what's making it lag. Hopefully i can actually get the site working again on my machine and take a look

hsheth2:
Yah we need to look into worksheet perf in general

@maxyuan6717 lmk if I can help with anything

maxyuan6717:
Aight lag should be gone. In general, you don't wanna pass props to components if it's not completely necessary to prevent rerenders, so I just moved the hover logic to css

frontend/src/components/WorksheetListItem.js
```diff
@@ -75,7 +75,7 @@ function WorksheetListItem({
           <span className={styles.course_title}>{course.title}</span>
         </Col>
         {/* Hide Button */}
-        <Col xs="auto" className={'pl-0 pr-2 my-auto hidden'}>
```
itsjoshthedeveloper: 😂

maxyuan6717: we don't talk about that

itsjoshthedeveloper: lol i was laughing at the commit name, typescript/eslint/prettier be making errors out of everything

itsjoshthedeveloper:
i made sure that the hide icon stays visible which the course is hidden and that the tooltip is placed at the bottom so it won't cover the course title

@hsheth2 do you know what's causing the php check to fail?

hsheth2:
The php check seems to be failing because it's started using php version 8 - you can ignore it here and I'll fix it in another branch

maxyuan6717 (approved):
fire

PR #455, opened 2021-01-10
Set php version in CI

hsheth2:


hsheth2:
cc @itsjoshthedeveloper - fyi

kevinhu (approved):

Issue #454, opened 2021-01-10
Fix catalog SearchResults component


itsjoshthedeveloper:
Turns out I broke something when I fixed the SearchResults stickiness in 9908ed7a58fba8d02df94d323f45c038a64f7e45...

Now the catalog search results are wonky when expanded and the card view on both the catalog and worksheet list view changed

PR #460, opened 2021-01-14
Fix padding of catalog stuff

maxyuan6717:


itsjoshthedeveloper (changes requested):
@maxyuan6717 the problems that i was talking about in #454 are shown in the gif below.

catalog-fix

more specifically, they are:
- the FB# gets pushed to the next row when coming from the worksheet
- the search results are not in the `SurfaceComponent` so the card view shows the cards without a background
- the expanded catalog view results don't have `mx-auto`, so they expand to the full width of the catalog

these are due to the changes found in 9908ed7a58fba8d02df94d323f45c038a64f7e45. i made these changes to fix the sticky headers for the search results but that resulted in the problems listed above.

maxyuan6717:
Ah interesting; Issues 1&3 don't happen on my screen width, so I missed them, but tis a quick fix

itsjoshthedeveloper (commented):
awesome, that looks great! the 1st issue was still there, so I decreased the width of the title column by 10px to prevent the FB# from overflowing.

itsjoshthedeveloper (approved):

itsjoshthedeveloper:
🙌

itsjoshthedeveloper:
@hsheth2 just noticed that you deployed the old version of `master`. this PR fixes the searchResults component bug, so can you deploy the branch after this merge instead?

hsheth2:
@itsjoshthedeveloper yep the latest master is now deployed

PR #571, opened 2021-03-03
Add whitelist to beta

itsjoshthedeveloper:


kevinhu (approved):
Looks good! One consideration: since we'll merge this into master eventually, we might want to have an env variable flag that disables the beta whitelist in production. That way, we can keep this whitelist functionality for future betas while disabling it on the main version.

itsjoshthedeveloper:
> Looks good! One consideration: since we'll merge this into master eventually, we might want to have an env variable flag that disables the beta whitelist in production. That way, we can keep this whitelist functionality for future betas while disabling it on the main version.

I was thinking about that and think it's a great idea, but wasn't sure how to implement it. How do you think we should do it?

maxyuan6717 (approved):

PR #609, opened 2021-03-13
About page update

nsnave:


nsnave:
Adding links for the members who want their website linked would be something straightforward for a new member to do. The TODO that mentions the links has been in the code for a while I think too lol.

maxyuan6717 (approved):
noice

PR #610, opened 2021-03-13
Yalies

kevinhu:
Yalies integration + changes to backend.
(not merging into master for now, since we still have to do some more testing)
- Grant account access based on Yalies info (check if student), delete challenge
- Merge API and reverse proxy into single route, and move frontend in front to facilitate Vercel deploys in future (along with required CORS settings for API)
- Configure API logging setup with winston+morgan

Test deployment of new API is available at api.coursetable.com. See alpha.coursetable.com for a preview version on Vercel that makes requests to this endpoint.


api/src/server.ts
```diff
@@ -0,0 +1,168 @@
...
+  for (const header of Object.keys(req.headers)) {
+    // Headers are automatically made lowercase by express.
+    if (header.startsWith('x-coursetable-')) {
+      delete req.headers[header];
+    }
+  }
+  next();
+});
+
+// Redirection routes for historical pages.
```
ErikBoesen, line 72: I don't know if there are any more of these, but if you want to make it easier to add new redirects you could just make an obj mapping endpoint to redirect URL and iterate through

docker/prod-compose.yml
```diff
@@ -110,41 +62,41 @@ services:
...
 
   php-fpm:
-    build: ./php
-    restart: always
-    env_file: ../facebook.env
+    # env_file: ../facebook.env
     environment:
+      FACEBOOK_APP_ID: ${FACEBOOK_APP_ID?fb_app_id}
+      FACEBOOK_APP_SECRET: ${FACEBOOK_APP_SECRET?fb_app_secret}
       MYSQL_HOST: ${MYSQL_HOST?host}
```
ErikBoesen, line 69: Out of curiosity what's the point of adding these defaults? Wouldn't it be a disabling issue if one of these env vars are undefined?

frontend/src/components/FerryProvider.tsx
```diff
@@ -93,21 +95,25 @@ const addToCache = (season: Season): Promise<void> => {
...
+    return axios
+      .get(`${API_ENDPOINT}/api/static/catalogs/${season}.json`, {
+        withCredentials: true,
+      })
+      .then((res) => {
+        // Convert season list into a crn lookup table.
+        const data = res.data as Listing[];
+        const info = new Map<Crn, Listing>();
+        for (const raw_listing of data) {
+          const listing = preprocess_courses(raw_listing);
```
ErikBoesen, line 107: I love snake case as much as the next python programmer, but why break the JS camel case standard only here?

api/src/auth/cas_auth.routes.ts
```diff
@@ -1,53 +1,111 @@
...
+              {
+                headers: {
+                  Authorization: `Bearer ${YALIES_API_KEY}`,
+                  'Content-Type': 'application/json',
+                },
+              }
+            )
+            .then(({ data }) => {
+              // if no user found, do not grant access
+              if (data === null || data.length === 0 || data.school) {
```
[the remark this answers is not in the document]
kevinhu: Ah my bad – that's an artifact from when I was checking the user instead.

api/src/auth/cas_auth.routes.ts
```diff
@@ -1,53 +1,111 @@
...
+    });
+  });
+};
 
 const postAuth = (req: express.Request, res: express.Response): void => {
   const redirect = req.query['redirect'] as string | undefined;
   if (redirect && !redirect.startsWith('//')) {
     if (redirect.startsWith('/')) {
-      return res.redirect(redirect);
+      return res.redirect(`${FRONTEND_ENDPOINT}${redirect}`);
```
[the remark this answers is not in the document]
kevinhu: Good idea!

api/src/auth/cas_auth.routes.ts
```diff
@@ -66,6 +124,39 @@ const casLogin = function (
...
+    req.headers['x-coursetable-authd'] = 'true';
+    req.headers['x-coursetable-netid'] = req.user.netId;
+
+    return next();
+  }
+  return next(new Error('CAS auth but no user'));
+};
+
+// middleware function for requiring user account + access to evaluations
+export const authHard = (
```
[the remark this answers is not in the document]
kevinhu: Sounds good – just an artifact from our previous system on the reverse proxy.

api/src/config.ts
```diff
@@ -1,27 +1,62 @@
...
+  : process.env.FRONTEND_ENDPOINT;
 
-export const NUM_CHALLENGE_COURSES = 3; // number of courses to select for the challenge
-export const CHALLENGE_SEASON = '201903'; // season to select the challenge from
-export const MAX_CHALLENGE_REQUESTS = 100; // maximum number of allowed challenge tries
+// CORS options so frontend can interface with server
+export const CORS_OPTIONS = {
+  origin: FRONTEND_ENDPOINT,
+  credentials: true,
+  optionsSuccessStatus: 200, // some legacy browsers (IE11, various SmartTVs) choke on 204
```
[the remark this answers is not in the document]
kevinhu: Haha not really my idea, just found it in the CORS tutorial I was looking at. Guess it can't hurt 🙃

api/src/server.ts
kevinhu: Yeah so these are the only ones for sure – we retain these for SEO purposes only.

docker/prod-compose.yml
kevinhu: Yep, this doesn't add a default but instead specifies a label. For instance, running without it specifies gives an error output:
```
Missing mandatory value for "environment" option interpolating {'FACEBOOK_APP_ID': '${FACEBOOK_APP_ID?fb_app_id}', 'FACEBOOK_APP_SECRET': '${FACEBOOK_APP_SECRET?fb_app_secret}'} in service "php-fpm": fb_app_id
```

frontend/src/components/FerryProvider.tsx
kevinhu: A leftover from our existing script, but it's worth changing

api/src/auth/cas_auth.routes.ts
```diff
@@ -1,53 +1,111 @@
...
+    return done(null, user.netId);
+  });
 
-passport.deserializeUser(function (netId: string, done) {
-  const user: User = {
-    netId,
-  };
-  done(null, user);
-});
+  // when deserializing, ping Yalies to get the user's profile
```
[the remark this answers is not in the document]
kevinhu: So this is used every time a user logs in – which is not very often, given that our cookies expire every year.

docker/prod-compose.yml
ErikBoesen: Makes sense, thank you!

api/src/auth/cas_auth.routes.ts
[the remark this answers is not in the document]
ErikBoesen: Makes sense @kevinhu

api/src/auth/cas_auth.routes.ts
```diff
@@ -88,13 +88,14 @@ export const passportConfig = (passport: passport.PassportStatic): void => {
...
   if (redirect && !redirect.startsWith('//')) {
-    if (redirect.startsWith('/')) {
-      return res.redirect(`${FRONTEND_ENDPOINT}${redirect}`);
+    // prefix the redirect with a slash to avoid an open redirect vulnerability.
+    if (!redirect.startsWith('/')) {
+      redirect = `/${redirect}`;
     }
-    // We prefix this with a slash to avoid an open redirect vulnerability.
-    return res.redirect(`${FRONTEND_ENDPOINT}/${redirect}`);
+    //
```
ErikBoesen, line 97: Remove

api/src/config.ts
```diff
@@ -14,34 +18,32 @@ export const SECURE_PORT = process.env.SECURE_PORT || 4096;
...
 
 // MySQL config for courses database (just worksheets)
 export const MYSQL_COURSES_CONFIG = {
-  host: process.env.MYSQL_HOST || die('mysql host'),
-  port: parseInt(process.env.MYSQL_PORT || die('mysql port'), 10),
-  user: process.env.MYSQL_USER || die('mysql username'),
-  password: process.env.MYSQL_PASSWORD || die('mysql password'),
-  database: process.env.COURSES_DB || die('courses db'),
+  host: getEnv('MYSQL_HOST'),
+  port: parseInt(getEnv('MYSQL_PORT'), 10),
```
ErikBoesen, line 36: Do you really need to specify base 10?

ErikBoesen (approved):

api/src/config.ts
kevinhu: Yeah, I think it's an eslint autofix: https://eslint.org/docs/rules/radix

ErikBoesen: Seems unnecessary, but ok.

PR #662, opened 2021-04-03
Worksheet for beta

itsjoshthedeveloper:


itsjoshthedeveloper:
don't merge just yet! i'm working on a tutorial for the worksheet too

frontend/src/searchContext.tsx
```diff
@@ -96,18 +105,38 @@ type Store = {
 const SearchContext = createContext<Store | undefined>(undefined);
 SearchContext.displayName = 'SearchContext';
 
+// Calculate upcoming season
+const dt = DateTime.now().setZone('America/New_York');
```
kevinhu, line 109: Should we also handle summer courses here?

frontend/src/searchContext.tsx
```diff
@@ -31,6 +38,8 @@ export type Option = {
   numeric?: boolean;
 };
 
+export const isOption = (x: any): x is Option => 'label' in x && 'value' in x;
```
kevinhu, line 41: Why do we need `x: any` followed by `x is Option`  here? Can we use a type besides `any`?

kevinhu (approved):
Looks good – this diff is way too large for me to have a sense of any architectural issues but it otherwise checks out.

frontend/src/searchContext.tsx
itsjoshthedeveloper: i thought about that but assumed most people don't do summer session. think we should?

frontend/src/searchContext.tsx
itsjoshthedeveloper: this is a type guard that we can use to check if any value is of type Option

PR #683, opened 2021-04-10
PostHog backend analytics

kevinhu:
Adds PostHog analytics to backend events. Because we are integrated with Yalies.io, this means we also track and store info such as year, residential college, graduation, etc., allowing us to get much more detailed user analytics than before (we only have NetIDs from frontend analytics).

Also mitigates the issue of PostHog being blocked by certain adblockers.

### DIFF
--- a/api/src/canny/canny.handlers.ts
+++ b/api/src/canny/canny.handlers.ts
@@ -97,6 +97,11 @@ export const cannyIdentify = async (
         },
       });
 
+      POSTHOG_CLIENT.capture({
+        distinctId: netId,
+        event: 'request-canny-token',
+      });
+
       const token = createCannyToken({
         netId,
         evals: !!user.school_code,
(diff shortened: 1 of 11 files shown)


achrysaetos (approved):

api/src/canny/canny.handlers.ts
```diff
@@ -81,6 +81,27 @@ export const cannyIdentify = async (
...
+        distinctId: netId,
+        properties: {
+          email: user.email || '',
+          firstName: user.first_name || '',
+          lastName: user.last_name || '',
+          upi: user.upi || -1,
+          school: user.school || '',
+          college: user.college || '',
+          major: user.major || '',
+          curriculum: user.curriculum || '',
```
ErikBoesen, line 96: Not sure, but you should be able to omit the default strings.

ErikBoesen (approved):

api/src/canny/canny.handlers.ts
kevinhu: Yeah I gave that a shot, but PostHog doesn't seem to like `undefined` or `null` values for these

PR #739, opened 2021-04-24
GraphiQL page

kevinhu:
- Adds a GraphiQL page for people to play around with constructing queries.
- Makes all requests to the Ferry backend as a Hasura read-only user (so people can't modify courses data)

What the editor looks like:
Screen Shot 2021-04-23 at 23 04 35


ErikBoesen (approved):
LGTM!

PR #786, opened 2021-05-07
Add beta announcement

itsjoshthedeveloper:


kevinhu (approved):
🚀

frontend/src/App.js
```diff
@@ -28,21 +28,32 @@ import { useUser } from './user';
...
+      Promise.allSettled([a, b]).finally(() => {
+        // Set loading to false after user info and fb info is fetched
+        setTimeout(setLoading, 3000, false);
+      });
+    }
+  }, [userRefresh, fbRefresh, firstLoad]);
+
+  // If user is a first-year, redirect to beta
+  if (user.year === 2025) {
+    window.location.href = `https://beta.coursetable.com${window.location.pathname}`;
```
itsjoshthedeveloper, line 55: idk if this is the best way to redirect, @kevinhu @maxyuan6717 what do you think?

kevinhu: Not sure about this, maybe give react-router a shot? https://reactrouter.com/web/api/Redirect

Also, have you confirmed that this works by setting your own year to 2025 and visiting?

itsjoshthedeveloper: yup it works! it seems like react router's Redirect component only works with paths on the same hostname URL..

itsjoshthedeveloper: switched it to `window.open()`

kevinhu: Just realized that there's another caveat here – if we only go by user year, we also include a bunch of grad school students as well as YC class of 2024 students on leave. Can you add conditions that the user has to be a YC student (should be the `school_code` attribute) and class of 2025 not on leave?

itsjoshthedeveloper: Yeah sure i can filter out the grad students, but should we filter out the class of 24/25 on leave? since they prob havent used the main coursetable much either. also don’t know how we would filter them out..

frontend/src/App.js
```diff
@@ -61,7 +66,7 @@ function App({ themeToggler }) {
   const MyRoute = Route;
 
   // Render spinner if page loading
-  if (loading) {
+  if (loading || user.year === 2025) {
```
ErikBoesen, line 69: Be warned: Yale is not currently listing years for 2025 students in the directory. They simply have no year listed, and unfortunately there's no way on the Yalies side to distinguish them from other people with no year listed.

itsjoshthedeveloper: Yep @kevinhu mentioned that. Since we're deploying the beta in june as the main version, it may not even matter. just thought I should include it in the case that Yale lists their year by June

frontend/src/components/CourseModal.js
```diff
@@ -105,7 +105,7 @@ const CourseModal = ({ listing, hideModal, show }) => {
   // key variable for lists
   let key = 0;
 
-  console.log(cur_listing);
+  // console.log(cur_listing);
```
kevinhu, line 108: Just remove this?

kevinhu (approved):

PR #1098, opened 2021-11-12
Fix same-courses queries and show past syllabi

kevinhu:
Fixes how we find past offerings of a course and adds a collapsible section to view all past syllabi.

Screen Shot 2021-11-12 at 18 15 19


itsjoshthedeveloper:
How about when courses have different course codes across seasons?

kevinhu:
It's complicated and we do our best: https://github.com/coursetable/ferry/blob/master/docs/3_same_classes.md

PR #1315, opened 2023-08-14
Remove outdated Notice

ErikBoesen:
JankTable has shut down, so I figured it might be time for this to go?


bearsyankees:
Yeah good call

PR #1324, opened 2023-09-05
Change banner wording: work for -> work on

ErikBoesen:
Not to nitpick the header for a second time in a row, but this wording implied to me that working on CourseTable is a job, which may turn some people away.


bearsyankees:
Thanks Erik keep it coming :)

PR #1325, opened 2023-09-07
Fix "Hide discussion sections" button

matthew-siff:
The "hide discussion sections" button currently checks whether the class name is "Discussion Section." This change checks whether the Section field is all letters, which is only true of discussion sections.


bearsyankees:
LGTM! Congrats on your first PR. In the future, just make a new branch, no need to make a whole fork :)

PR #1326, opened 2023-09-09
Search in descriptions of courses if applicable

GoldinGuy:
A common pain point I've heard about Coursetable is that the search doesn't include content within course descriptions. In theory, this one line of code should add that.

### DIFF
--- a/frontend/src/contexts/searchContext.tsx
+++ b/frontend/src/contexts/searchContext.tsx
@@ -1,137 +1,137 @@
-import _ from 'lodash';
-import posthog from 'posthog-js';
+import _ from "lodash";
+import posthog from "posthog-js";
 import React, {
-  createContext,
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useState,
-} from 'react';
-import { GroupedOptionsType, OptionsType } from 'react-select/src/types';
+	createContext,
+	useCallback,
+	useContext,
+	useEffect,
+	useMemo,
+	useState,
+} from "react";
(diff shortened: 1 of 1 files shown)


frontend/src/contexts/searchContext.tsx
```diff
@@ -148,907 +148,908 @@ SearchContext.displayName = 'SearchContext';
...
+					listing.number.toLowerCase().startsWith(token) ||
+					// for course numbers that start with a letter
+					// (checked by if .toLowerCase() is not equal to .toUpperCase(), see https://stackoverflow.com/a/32567789/5540324),
+					// exclude this letter when comparing with the search token
+					(numberFirstChar.toLowerCase() !== numberFirstChar.toUpperCase() &&
+						listing.number
+							.toLowerCase()
+							.startsWith(numberFirstChar.toLowerCase() + token)) ||
+					listing.title.toLowerCase().includes(token) ||
+					listing.description?.toLowerCase()?.includes(token) ||
```
thepolitic, line 728: this is the line we want, right? Would it be possible to just commit this without the rest of the formatting changes? and maybe add it as a button with toggle on (if not I/someone else can work on doing that, shouldn't be hard).

bearsyankees, line 728: this is the line we want, right? Would it be possible to just commit this without the rest of the formatting changes? and maybe add it as a button with toggle on (if not I/someone else can work on doing that, shouldn't be hard).

bearsyankees (changes requested):
looks solid just change formatting ideally and maybe integrate to button (or someone/i can)

PR #1338, opened 2023-10-04
Cleanup Part 1

neilsong:
Bumped all API dependencies to latest versions (with the exception of passport 0.6 cuz its goofy and doesn't work). Also rewrote some promise handling from .then/.catch to await with types so TS compiles.

Oh and formatted everything in API with eslint-prettier.

Edit 1: Prisma also updated to latest.


api/src/catalog/catalog.d.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @file User type declarations.
+ */
```
Josh-Cena, line 3: Don't use `.d.ts` for type declarations—use a normal `.ts` file instead. Also where were these types before? Did they not exist?

Josh-Cena (changes requested):

api/src/catalog/catalog.d.ts
neilsong: Done :)

We can clean up explicitly defined types after you implement the runtime lib.

neilsong (commented):
Implemented Sida's type suggestion

Josh-Cena:
Do you know why frontend/yarn.lock is updated? IIUC they are separate projects and bumping one shouldn't update the other

api/src/catalog/catalog.utils.ts
```diff
@@ -9,14 +9,72 @@ import { request } from 'graphql-request';
 
 import winston from '../logging/winston';
 
+/**
+ * Type for Catalog data
+ */
```
Josh-Cena, line 14: Extraneous comment?

api/src/catalog/catalog.utils.ts
```diff
@@ -9,14 +9,72 @@ import { request } from 'graphql-request';
...
+
+export type SeasonsType = {
```
Josh-Cena, line 16: I would recommend removing the `Type` suffix. You wouldn't say `seasonValue` either.

Josh-Cena (commented):
Incomplete reviews (the challenge controllers needs a deeper look); just some nits. Overall LGTM

maxyuan6717:
neil i stg if u break coursetable with this

neilsong:
yo @maxyuan6717 chilllll, have some faith in lil bro 🙏

@Josh-Cena, will double check the frontend yarn.lock and the challenge workflow.

api/src/catalog/catalog.utils.ts
neilsong: so just `Seasons` and `Catalog`?

api/src/catalog/catalog.utils.ts
```diff
@@ -9,14 +9,72 @@ import { request } from 'graphql-request';
...
+  seasons: {
+    season_code: string;
+    term: string;
+    year: number;
+  }[];
+};
+
+// type for catalog data
```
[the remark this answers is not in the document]
neilsong: mhm

bearsyankees:
Looking over this now. @Josh-Cena thank you for also helping with this review. @maxyuan6717 blame neil if it breaks

api/src/catalog/catalog.utils.ts
Josh-Cena: right

neilsong:
Decided to include the prisma migration in this PR since there are no breaking changes with latest.

api/src/challenge/challenge.controllers.ts
```diff
@@ -149,8 +149,10 @@ export const requestChallenge = async (
   // randomly choosing a minimum rating
   const minRating = 1 + Math.random() * 4;
 
+  let evals: requestEvalsQueryResponse;
+  // get a list of all seasons
   try {
-    const evals = await request(GRAPHQL_ENDPOINT, requestEvalsQuery, {
+    evals = await request(GRAPHQL_ENDPOINT, requestEvalsQuery, {
```
Josh-Cena, line 155: If it's only used inside the block, put it inside the block

api/src/challenge/challenge.controllers.ts
```diff
@@ -291,14 +293,14 @@ export const verifyChallenge = async (
     });
   }
 
+  let true_evals: verifyEvalsQueryResponse;
+  // get a list of all seasons
   try {
-    // check the answers against the true values
-    const trueEvals = await request(GRAPHQL_ENDPOINT, verifyEvalsQuery, {
+    true_evals = await request(GRAPHQL_ENDPOINT, verifyEvalsQuery, {
```
Josh-Cena, line 299: Use camelCase + move inside the block

Josh-Cena (approved):
Almost LGTM

Josh-Cena (approved):
Great!

bearsyankees (approved):
#CourseTableIsBack

PR #1340, opened 2023-10-04
chore: Bun package manager and CI 🍞 

neilsong:


Josh-Cena:
Reminder that you need to update the docs as well

neilsong:
> Reminder that you need to update the docs as well. Also let's use the same package manager for frontend–don't make them out of sync. Also how good is Bun's support for monorepos?

Will do on the frontend. Not sure about the monorepo support tbh - we can find out :)

Josh-Cena:
I think it does: https://bun.sh/docs/install/workspaces It's probably not much less suckier than npm workspaces but our monorepo is small anyway, so I'm +1 on this.

bearsyankees (dismissed):
LGTM just fix the conflicts and can ship

api/package.json
```diff
@@ -1,11 +1,12 @@
 {
   "scripts": {
     "prestart": "yarn prisma db pull && yarn prisma generate",
+    "generate": "bun prisma generate",
     "start": "yarn ts-node-dev --debug --files --exit-child src/server.ts",
     "prod": "yarn ts-node --files src/server.ts",
```
Josh-Cena, line 6: Should these also use `bun`? Also why is `generate` now separate?

Josh-Cena:
I can still see a few occurrences of `yarn` in numerous files; maybe update those as well?

api/package.json
neilsong: For some reason it's necessary with bun's install - I think it's cuz it doesn't yarn's postinstall prisma client build step.

neilsong:
> I can still see a few occurrences of `yarn` in numerous files; maybe update those as well?

Can't run things with current bun support, only package management and ci tasks work with prisma+bun. Are there any others that you're referring to that aren't running/building?

Josh-Cena:
So we are still using a mix of bun and Yarn? In this case we need to keep the yarn.lock because they are still used.

neilsong:
Bun for dependencies, Yarn for running/building. I tested it - `yarn run` works without a yarn.lock file

Josh-Cena:
Ah I see. Fair enough then

neilsong:
But also we could completely migrate to Bun + `npm run` and phase out Yarn. It doesn't seem to offer much value with this setup to be honest.

Josh-Cena (commented):
Here are the remaining mentions of Yarn I found:

- In `api/Dockerfile`, we are still referencing yarn.lock and `yarn install`/`yarn start`
- In `docker/docker-compose.yml`, we are still using `yarn start` (which may be intentional)
- In `docs/getting-started.md` on line 217 (which I can't comment on in the diff view), we are still using `yarn add`
- `frontend/README.md` exclusively talks about Yarn
- `frontend/start.sh` still uses `yarn start` (again, which may be intentional)
- In `frontend/upgrade-packages.sh`, we are still using `yarn upgrade` (this script should be deleted I guess. @bearsyankees?)

If using Yarn is intentional in any of these places because Bun doesn't work, I would add a `TODO` comment explaining why as we should eventually be exclusively on Bun.

neilsong:
> Here are the remaining mentions of Yarn I found:
> 
> * In `api/Dockerfile`, we are still referencing yarn.lock and `yarn install`/`yarn start`
> * In `docker/docker-compose.yml`, we are still using `yarn start` (which may be intentional)
> * In `docs/getting-started.md` on line 217 (which I can't comment on in the diff view), we are still using `yarn add`
> * `frontend/README.md` exclusively talks about Yarn
> * `frontend/start.sh` still uses `yarn start` (again, which may be intentional)
> * In `frontend/upgrade-packages.sh`, we are still using `yarn upgrade` (this script should be deleted I guess. @bearsyankees?)
> 
> If using Yarn is intentional in any of these places because Bun doesn't work, I would add a `TODO` comment explaining why as we should eventually be exclusively on Bun.

Will take a look at all these and add comments tomorrow. As for the Docker stuff, I will open another PR to rework that whole workflow.

Josh-Cena:
Thanks! I think the Docker rework should be a prereq for this. (All of the sh*t gets entangled and so hard to pull apart, eh?)

bearsyankees:
> Here are the remaining mentions of Yarn I found:
> 
> * In `api/Dockerfile`, we are still referencing yarn.lock and `yarn install`/`yarn start`
> * In `docker/docker-compose.yml`, we are still using `yarn start` (which may be intentional)
> * In `docs/getting-started.md` on line 217 (which I can't comment on in the diff view), we are still using `yarn add`
> * `frontend/README.md` exclusively talks about Yarn
> * `frontend/start.sh` still uses `yarn start` (again, which may be intentional)
> * In `frontend/upgrade-packages.sh`, we are still using `yarn upgrade` (this script should be deleted I guess. @bearsyankees?)
> 
> If using Yarn is intentional in any of these places because Bun doesn't work, I would add a `TODO` comment explaining why as we should eventually be exclusively on Bun.



> Here are the remaining mentions of Yarn I found:
> 
> * In `api/Dockerfile`, we are still referencing yarn.lock and `yarn install`/`yarn start`
> * In `docker/docker-compose.yml`, we are still using `yarn start` (which may be intentional)
> * In `docs/getting-started.md` on line 217 (which I can't comment on in the diff view), we are still using `yarn add`
> * `frontend/README.md` exclusively talks about Yarn
> * `frontend/start.sh` still uses `yarn start` (again, which may be intentional)
> * In `frontend/upgrade-packages.sh`, we are still using `yarn upgrade` (this script should be deleted I guess. @bearsyankees?)
> 
> If using Yarn is intentional in any of these places because Bun doesn't work, I would add a `TODO` comment explaining why as we should eventually be exclusively on Bun.

Yea, can delete it, i just made it to be helpful for seeing which packages would break everything when trying to upgrade all. It's not used just a helper function for local dev

bearsyankees:
> Thanks! I think the Docker rework should be a prereq for this. (All of the sh*t gets entangled and so hard to pull apart, eh?)

Agreed

neilsong:
Actually @Josh-Cena @bearsyankees - I thought about it some more: I think we should go ahead and merge this for using Bun in installs and CI. I will rewrite the docs references (changing `yarn add/run` to `npm install/run` and migrate to `bun run` and/or `npm run` from `yarn start` in a separate Docker PR. 

`bun run` and the current Docker config are a bit dodgy so I'd rather get that resolved in a separate PR and ship these Bun CI + package management features that work independently of Docker. What are your thoughts?

Josh-Cena:
My thought is that we should not have the repo in an "intermediate state"—it should be production-ready at every master commit. If we merge this one, then we would be in an "intermediate state" where the docs, docker script, and npm script are out of sync. Therefore I think this has a hard dependency on the Docker refactor.

Josh-Cena (approved):
Let's do this to unblock you from the docker refactor

PR #1344, opened 2023-10-05
chore: update dev deps in frontend

Josh-Cena:
First step to clean the frontend.


api/src/config.ts
bearsyankees, line 1: ik u don't like to use 3000 but lets leave it as that for others i think maybe, see my slack

frontend/src/App.jsx
bearsyankees, line 1: nice

bearsyankees, line 1: good commas.

api/src/config.ts
Josh-Cena: Okay, I've reverted back to 3000 by default.

frontend/src/App.jsx
Josh-Cena: FYI, we used to have a conflict because formatOnSave uses the latest Prettier which automatically adds these commas, but CI uses v2 which removes them. So this should probably be prioritized for that reason too

api/src/config.ts
```diff
@@ -37,16 +37,20 @@ export const MAX_CHALLENGE_REQUESTS = 100; // maximum number of allowed challeng
 // Secret for Canny SSO
 export const CANNY_KEY = getEnv('CANNY_KEY');
 
+// TODO make sure the frontend port is synchronized with this;
+// we should use the same script to start both frontend and backend
+const port = process.env.FRONT_END_PORT || 3000;
```
Josh-Cena, line 42: FYI, I've added a `PORT` env to frontend and `FRONT_END_PORT` to api.

bearsyankees: great

bearsyankees (approved):
looks good!

PR #1345, opened 2023-10-06
chore: upgrade non-breaking deps in frontend

Josh-Cena:
Part 2 of frontend cleanup. This one handles all the probably-non-breaking upgrades and leaves the rest to future PRs.


bearsyankees (approved):
more packages yay nice

PR #1349, opened 2023-10-06
chore: display the right host in dev

bearsyankees:
@Josh-Cena what do ya think


Josh-Cena (approved):
Other than formatting issues LGTM

PR #1351, opened 2023-10-07
chore: Docker refactor 🐳

neilsong:

PR #1352, opened 2023-10-07
Hotfix API and Add NPM Compatibility

neilsong:
This works for sure.


neilsong:
Ok @Josh-Cena it says it fails to deploy and frontend fails but like it definitely works, just look at the `last of npm compat` commit in #1351

PR #1353, opened 2023-10-07
add description search feature

CilantroMalik:
First draft of description search feature; I did some preliminary testing and I think this works but feedback is welcome (yay first feature!)


GoldinGuy:
My PR was closed in favor of this PR (which is great as this is better fleshed out and closer to completion). However, a critical component that is different (based on a cursory overview of the changes) is that this PR adds searching the description as an additional, optional feature rather than making that the default. I feel strongly that searching the description should be the default functionality – Coursetable should surface *more* content by default, and then filters cut down on that.

GoldinGuy:
In response to [this comment](https://github.com/coursetable/coursetable/pull/1329#pullrequestreview-1662350140), descriptions do contain arbitrary text, but it's ultimately text related to the course. The point of this addition (in my original understanding) was to implement an almost "semantic search" like functionality into search by surfacing courses where the key term might not necessarily be in the title. 

While adding this an as optional checkbox might have essentially the same effect for power users like you or me, the reality is that the average user won't know to enable this checkbox and the majority of searches will continue taking place without searching the description.

GoldinGuy:
My PR was integrated with this one, which is excellent as it's more detailed and nearly complete. But – an observation I had is that this PR offers searching the description as an option rather than a default. I believe having the description search as the default could be extremely beneficial – Coursetable can show more content initially, and then filters can be used to narrow it down.

Regarding [this comment](https://github.com/coursetable/coursetable/pull/1329#pullrequestreview-1662350140), while descriptions might have varied text, it's still relevant to the course. My initial thought was to enhance the search to be more "semantic-like" and highlight courses where the key term might not be in the title.

Having an optional checkbox is great for experienced users, but the reality is that the average student won't know to enable this checkbox and the majority of searches will continue taking place without searching the description.

bearsyankees:
> My PR was integrated with this one, which is excellent as it's more detailed and nearly complete. But – an observation I had is that this PR offers searching the description as an option rather than a default. I believe having the description search as the default could be extremely beneficial – Coursetable can show more content initially, and then filters can be used to narrow it down.
> 
> Regarding [this comment](https://github.com/coursetable/coursetable/pull/1329#pullrequestreview-1662350140), while descriptions might have varied text, it's still relevant to the course. My initial thought was to enhance the search to be more "semantic-like" and highlight courses where the key term might not be in the title.
> 
> Having an optional checkbox is great for experienced users, but the reality is that the average student won't know to enable this checkbox and the majority of searches will continue taking place without searching the description.

Yes good points. This will come soon

bearsyankees:
NICE JOB!!!!

bearsyankees:
thank you so much @GoldinGuy

benzuzu:
awesome stuff @GoldinGuy

Etherite1:
thanks @GoldinGuy

GoldinGuy:
Looks great guys! Definitely a bit hard to find especially on mobile but this works really well - very cool to be able to search terms like "War" or "ML" and have all relevant courses appear :)

PR #1347, opened 2023-10-06
chore: React 18, react-router 6

Josh-Cena:
@bearsyankees Please help me test this—there could be breaking changes I didn't catch.

PR #1361, opened 2023-10-09
refactor: preparatory work for React 18 upgrade

Josh-Cena:
This comes out of https://github.com/coursetable/coursetable/pull/1347. It contains all non-breaking refactors so is hopefully easier to review. Major changes:

1. Removed some excess `useCallback` that essentially doesn't do much while being weakly typed
2. Used `useLocation` instead of `withRouter`
3. Got rid of `React.FC`—it's discouraged in the community
4. Other refactors to get around type errors


bearsyankees (changes requested):
check the new conflicts

PR #1363, opened 2023-10-10
fix: Docker hotfix for prod

neilsong:
Fix catalogs error and also remove redundant windows compat.


api/prod-compose.yml
```diff
@@ -1,10 +1,14 @@
...
     restart: always
+    ports:
+      - '8039:8039'
+      - '8038:8080'
     expose:
-      - 8039
+      - '8039'
     environment:
       NODE_ENV: production
+      OVERWRITE_CATALOG: 'true'
```
Josh-Cena, line 11: Does it not work without? This thing seems to be the route handler, which means it's run on every request, right?

Josh-Cena (commented):
LGTM, with one uncertainty

api/prod-compose.yml
neilsong: You're right, it works without since the container is recreated with a new static every time. I will add overwrite catalog as a bash argument to ./start.sh in dev since there may be times where we want to update catalogs in the mounted filesystem approach in dev.

Josh-Cena:
If we are fully onboard to `bun`, can we get rid of the rest of the mentions for `npm` (`yarn` seems gone now)?

PR #1367, opened 2023-10-10
chore: rework issue/PR templates, remove dependabot

Josh-Cena:
We don't actually use dependabot and the amount of PRs it creates is annoying.

The PR template is not being picked up; I also took the chance to make the templates nicer.


bearsyankees (approved):
🐐

PR #1382, opened 2023-11-11
chore: Dockerfile Refactor for Prisma Fix

neilsong:
Fixes: 
 - The Dockerfile for api has been refactored to fix the `prisma` client generation from scratch. This also works on latest `prisma`

Features:
 - The Dockerfile to a multi-stage build
   1.  `bun` install (note that open-ssl is 1.x)
   2. `npm` generate (note that open-ssl isn't installed, so prisma defaults to building for 1.x)
   3. `ts-node-dev` entry with hot reloading
 - Dev and prod can use the same Dockerfile (much more readable)

## Edit 2
 - Use `ts-node-dev` over `nodemon`
 - Use `ts-node` in prod 

## Edit 1
 - `bun` is not feasible as an entry (updated features section to reflect current config)
   - This is because `prisma` cannot run `db pull` with `bun` runtime


api/package.json
```diff
@@ -46,8 +43,7 @@
     "morgan": "^1.10.0",
     "passport": "0.5.0",
     "passport-cas": "git+https://github.com/coursetable/passport-cas",
-    "prisma": "5.4.0",
-    "ts-node": "^10.9.1",
+    "prisma": "^5.2.0",
```
Josh-Cena, line 46: Check node_modules and see what version this resolves to? `^5.2.0` should match the latest version. I don't know if that's the intention; if so, you should directly use `^5.5.2`.

neilsong: I mean sure. Eventually, it'll always be outdated when 5.5.3 or 5.6.0 comes out - that's why we use the carrot semver right?

Josh-Cena: Indeed. My point is you used a lower range that resolves to a higher version which seems confusing. If this is intended, feel free to not do anything, although I personally bump the lower bound whenever I actually mean to bump it.

neilsong: Gotcha. Yea didn't really mean to bump it in this pq but I've updated the ver anyways.

PR #1388, opened 2023-11-13
chore: add sc to About

Josh-Cena:


neilsong (approved):
W PR as always, just doing some PR cleaning before break hits.

PR #1402, opened 2023-11-16
Multiple worksheets

wesleyandrade05:
## Pre-flight checklist

- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Summary

This PR fixes/implements the following **bugs/features**

- Add worksheets
- Delete worksheets
- Rename worksheets

## Test plan (required)

Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.

## Related issues/PRs

Fixes #

### DIFF
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -62,7 +62,7 @@ model WorksheetNames {
   worksheet_number Int    @db.UnsignedMediumInt
   name             String @db.Char(255)
 
-  @@unique([net_id, worksheet_number, name], map: "net_id_worksheet_number_name")
+  @@unique([net_id, worksheet_number], map: "net_id_worksheet_number")
   @@index([net_id], map: "net_id")
 }
 
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -57,6 +57,15 @@ model WorksheetCourses {
   @@index([net_id], map: "net_id")
 }
 
+model WorksheetNames {
+  net_id           String @db.Char(8)
+  worksheet_number Int    @db.UnsignedMediumInt
+  name             String @db.Char(255)
+
+  @@unique([net_id, worksheet_number, name], map: "net_id_worksheet_number_name")
+  @@index([net_id], map: "net_id")
+}
+
 model SavedCourses {
   id          Int    @id @default(autoincrement()) @db.UnsignedMediumInt
   net_id      String @db.Char(8)
--- a/frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
+++ b/frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
@@ -1,6 +1,12 @@
 /* eslint-disable guard-for-in */
 import React, { useMemo, useState } from 'react';
-import { Form, Row, ToggleButton, ToggleButtonGroup, Button } from 'react-bootstrap';
+import {
+  Form,
+  Row,
+  ToggleButton,
+  ToggleButtonGroup,
+  Button,
+} from 'react-bootstrap';
 import styled from 'styled-components';
 import { ValueType } from 'react-select/src/types';
 import { components } from 'react-select';
@@ -109,7 +115,7 @@ export function NavbarWorksheetSearch() {
     if (worksheet_number) {
       return {
         value: worksheet_number,
-        label: worksheet_options[parseInt(worksheet_number)].label
+        label: worksheet_options[parseInt(worksheet_number)].label,
       };
     }
     return null;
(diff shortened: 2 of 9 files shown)


frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
```diff
@@ -189,6 +193,24 @@ export function NavbarWorksheetSearch() {
...
+      ? 'Reached maximum number of worksheets'
+      : 'Add worksheet',
+  );
+
+  const [boolWorksheetsNumber, setBoolWorksheetsNumber] = useState(
+    worksheet_options.length >= 20 ? true : false,
+  );
+
+  const [deleteBool, setDeleteBool] = useState(
+    worksheet_number in ['0', '1', '2', '3'] ? true : false,
```
Josh-Cena, line 211: Do you mean `['0', '1', '2', '3'].includes(worksheet_number)`?

frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
```diff
@@ -252,10 +274,92 @@ export function NavbarWorksheetSearch() {
                 placeholder="Main Worksheet"
                 onChange={(selectedOption: ValueType<Option, boolean>) => {
                   if (isOption(selectedOption)) {
+                    if (selectedOption.value in ['0', '1', '2', '3']) {
```
Josh-Cena, line 277: Again: don't use `in` on arrays. It doesn't do what you think

frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
```diff
@@ -252,10 +274,92 @@ export function NavbarWorksheetSearch() {
...
+                      setDeleteBool(true);
+                    } else {
+                      setDeleteBool(false);
+                    }
```
Josh-Cena, line 281: This entire thing is `setDeleteBool(['0', '1', '2', '3'].includes(selectedOption.value)`

frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
```diff
@@ -189,6 +193,24 @@ export function NavbarWorksheetSearch() {
...
+
+  const [newWorksheetName, setNewWorksheetName] = useState('');
+
+  const [addWorksheetPlaceHolder, setAddWorksheetPlaceHolder] = useState(
+    worksheet_options.length >= 20
+      ? 'Reached maximum number of worksheets'
+      : 'Add worksheet',
+  );
+
+  const [boolWorksheetsNumber, setBoolWorksheetsNumber] = useState(
```
Josh-Cena, line 206: Don't call a variable "bool". Call it something like `maxWorksheetsNumExceeded` or something like that

frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
```diff
@@ -189,6 +193,24 @@ export function NavbarWorksheetSearch() {
...
+    worksheet_options.length >= 20 ? true : false,
+  );
+
+  const [deleteBool, setDeleteBool] = useState(
```
Josh-Cena, line 210: Call it `deleted`?

api/prisma/schema.prisma
```diff
@@ -57,6 +57,15 @@ model WorksheetCourses {
...
 }
 
+model WorksheetNames {
+  net_id           String @db.Char(8)
+  worksheet_number Int    @db.UnsignedMediumInt
+  name             String @db.Char(255)
+
+  @@unique([net_id, worksheet_number], map: "net_id_worksheet_number")
+  @@index([net_id], map: "net_id")
+}
```
Josh-Cena, line 67: I don't know much about Prisma so I can't comment on this, but this looks weird to me: shouldn't the worksheets data structure just be one `name -> courses` map? Why do we do `name -> index -> courses`?

Josh-Cena (commented):
Incomplete reviews

bearsyankees:
> Thank you for taking on this feature!
> 
> image
> 
> I just have a few suggestions for some improvements to the overall ui/ux of the feature.
> 
> Approach 1:
> 
> * For renaming, do you think it'd better if you just make the active selection component (rounded outline at top) editable with a check mark on the side to confirm? It already has a cursor anyways.
> * You can also implement delete by adding a small grey "x" icon right-aligned to the list components.
> 
> OR
> 
> **Approach 2 (strong recommend):**
> 
> * Again make the active selection editable
> * Instead of using check and x, you can follow the similar button at bottom right of dropdown, creating a blue one for rename and a red one for delete (like you have currently). Probably just set the rename one as disabled by default if the active selection text value is the same as the worksheet name.
> 
> Both are just suggestions, but definitely do add some sort of confirmation/button for renaming action.
> 
> As for adding worksheets, I recommend creating another drop down with a + icon next to the light/dark icon that has an "add worksheet" option and using a dialog (modal-style). I think this warrants another PR though (because this is a major UI change, and we should put "add friend" in there as well), so feel free to just leave out adding worksheet functionality for now.
> 
> Also, I just realized we shouldn't let people delete all their worksheets without giving an add option, so just add rename functionality for now, and comment delete button out. Let's reactivate it in another small PR after the + icon PR is merged.
> 
> Do keep all the function logic though.
> 
> Edit 1: @bearsyankees @Josh-Cena Please feel free to share your own thoughts as well. I definitely think we should talk about the + icon approach in-depth, but I think this is much better for long-run modularity. This PR and #1384 are the first times we're adding user-controlled workflows for modifying higher-level containers that aren't classes.
> 
> Edit 2: Clarifying a half-baked sentence about the dialog: open the dialog on "add worksheet" option click and use a standard add text field + confirm button dialog setup.

Thanks for all these Neil. I agree, I think we definitely need a major UI rework on both of these dropdowns. Can get a designer on that. And we have some time before next registration, so this can get very fleshed out. I think the best option is to have the selection editable, with either rename or delete as the options as you've said. Great job @wesleyandrade05 and thanks for reviewing @neilsong @Josh-Cena

api/prisma/schema.prisma
Josh-Cena: After some more thought, I think this makes more sense.

Josh-Cena:
Hi, it has been a while and too much has changed. I'm going to close this for now because of the accumulating merge conflicts. You can create a new branch and re-apply the changes (but beware of the updated code style).

PR #1409, opened 2023-11-28
fix: display all class meet times

Josh-Cena:
## Pre-flight checklist

- [x] I have searched for potential duplicate pull requests.
- [x] **If this is a code change**: I have written unit tests and/or conducted manual tests to fully verify the new behavior.
- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes https://feedback.coursetable.com/bugs/p/see-all-meeting-times-for-a-class) and the maintainers have approved on my working plan.

## Summary

This PR makes all times and locations considered by:

1. Conflict checking
2. Calendar view
3. Export ICS/GCal
4. Course modal

## Test plan (required)

Before:

image

After:

image

Before:

image

After:

image

PR #1410, opened 2023-11-28
fix: display time in 12h in modal

Josh-Cena:
Follow up to https://github.com/coursetable/coursetable/pull/1409


bearsyankees (approved):
noice

PR #1412, opened 2023-11-28
fix: do not "fix" class time before 8

Josh-Cena:
We have classes that happen before 8am. These classes are wrongly "fixed" to happen in the afternoon. Fixes https://coursetable.canny.io/admin/feedback/bugs/p/early-classes-show-up-in-the-evening-on-the-calendar


bearsyankees (approved):
Makes sense good fix!

PR #1417, opened 2023-11-29
fix: URL Encode Branch Names for Preview URLs

### LINKED ISSUE #1416: Bug Fixes for Preview CD
Automatically rewrite the `/` and other subdomain URL-prohibited characters in branch names on Frontend Preview CD deploys.

neilsong:
Fixes #1416 

(also fixes frontend lints)

### DIFF
--- a/.github/workflows/preview_cd.yml
+++ b/.github/workflows/preview_cd.yml
@@ -11,7 +11,7 @@ jobs:
   frontend-build-and-deploy:
     environment: 
       name: Preview
-      url: https://${{ env.sha_short }}.${{ github.head_ref }}.preview.coursetable.com
+      url: https://${{ env.sha_short }}.${{ steps.encode_url_branch.outputs.result }}.preview.coursetable.com
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/preview_cd.yml
```diff
@@ -22,6 +22,13 @@ jobs:
         shell: bash
         run: |
            echo "sha_short=`echo ${{ github.event.pull_request.head.sha }} | cut -c1-8`" >> $GITHUB_ENV
+      - name: Encode URL Branch
+        id: encode_url_branch
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: |
+            return "${{ github.head_ref }}".replace(/[^0-9a-zA-Z\-\.]+/g, "-")
```
Josh-Cena, line 31: What if `head_ref` contains `"`?

Josh-Cena:
I've always been curious: what happens if multiple branches with the same name were created (not simultaneously)? Do the deployments get deleted once the PR is merged?

.github/workflows/preview_cd.yml
neilsong: I mean at that point, we need to have a talk about naming branches :sob:

Everything should follow the convention defined in `README` and this is really only so that dependabot can work. Even if I do the backtick literal for strings, the branch name could always have a backtick so I feel like this one's a bit unwinnable without complicating it further than it has to be.

neilsong:
Deployments don't get deleted but technically we can't have two branches of the same name on remote right? I don't think recycling should be an issue.

.github/workflows/preview_cd.yml
Josh-Cena: Yeah... Let's not worry too much about it now. Just that it's potentially hackable.

Josh-Cena:
Okay, so deployments can be recycled. Good to know.

PR #1418, opened 2023-11-29
feat: unique URL for each course modal

Josh-Cena:
Enables us to add links


Josh-Cena:
Yes, still overlay it on `worksheet`. I don't think it matters; there's no indication that it's _from_ the worksheet anyway.

PR #1420, opened 2023-11-29
fix: don't double count in worksheet stats

bearsyankees:
just a start


bearsyankees:
@Josh-Cena thoughts as a start

frontend/src/components/Worksheet/WorksheetStats.tsx
```diff
@@ -31,6 +31,9 @@ const workloadColormap = chroma
 export default function WorksheetStats() {
   const [shown, setShown] = useState(true);
   const { courses, hidden_courses, cur_season } = useWorksheet();
+  //console.log(courses);
+  const countedCourseCodes = new Map();
```
Josh-Cena, line 35: Use `Set`

frontend/src/components/Worksheet/WorksheetStats.tsx
```diff
@@ -39,18 +42,31 @@ export default function WorksheetStats() {
...
-            workload: acc.workload + (c.average_workload ?? 0),
-            rating: acc.rating + (c.average_rating ?? 0),
-            skillsAreas: [...acc.skillsAreas, ...c.skills, ...c.areas],
-          },
+    (acc, c) => {
+      //const sectionIsInt = Number.isInteger(parseInt(c.section));
+      // see if any of the course's codes have already been counted or if it's hidden so we don't double count
+      const alreadyCounted =
+        c.all_course_codes.some((code) => countedCourseCodes.get(code)) ||
+        hidden_courses[cur_season]?.[c.crn];
```
Josh-Cena, line 50: This is not `alreadyCounted`, it's `shouldNotCount` :)

Josh-Cena:
Also how do we make sure that the first course in the list is the one that has credit information?

bearsyankees:
> Also how do we make sure that the first course in the list is the one that has credit information?

Yes this still needs to be addressed, will add that

bearsyankees:
@Josh-Cena Ok does that logic make sense for making sure we only count the one w the relevant info?

bearsyankees:
wait maybe should be and not or

frontend/src/components/Worksheet/WorksheetStats.tsx
```diff
@@ -39,18 +42,32 @@ export default function WorksheetStats() {
...
-            rating: acc.rating + (c.average_rating ?? 0),
-            skillsAreas: [...acc.skillsAreas, ...c.skills, ...c.areas],
-          },
+    (acc, c) => {
+      //const sectionIsInt = Number.isInteger(parseInt(c.section));
+      // see if any of the course's codes have already been counted or if it's hidden so we don't double count
+      const shouldNotCount =
+        c.all_course_codes.some((code) => countedCourseCodes.has(code)) ||
+        hidden_courses[cur_season]?.[c.crn];
+      const hasCourseInfo = c.average_rating && c.credits && c.average_workload;
```
Josh-Cena, line 51: If a course has no rating but has credits, we should count it nontheless, no?

bearsyankees: Ok so the scenario is just when it has no credits skip, otherwise treat it as the one to get data from?

Josh-Cena: Yeah that sounds reasonable

frontend/src/components/Worksheet/WorksheetStats.tsx
```diff
@@ -39,18 +42,32 @@ export default function WorksheetStats() {
...
+      const shouldNotCount =
+        c.all_course_codes.some((code) => countedCourseCodes.has(code)) ||
+        hidden_courses[cur_season]?.[c.crn];
+      const useCourseInfo =  c.credits;
+
+      if (shouldNotCount || !useCourseInfo) {
+        return acc;
+      }
+
+      // mark codes as counting, no double counting
```
Josh-Cena, line 57:
```suggestion
      // Mark codes as counted, no double counting
```

Josh-Cena (approved):
Remove the logs and then LGTM

frontend/src/components/Worksheet/WorksheetStats.tsx
```diff
@@ -39,18 +42,32 @@ export default function WorksheetStats() {
...
-            courseCnt: acc.courseCnt + 1,
-            coursesWithRating:
-              acc.coursesWithRating + (c.average_rating ? 1 : 0),
-            credits: acc.credits + (c.credits ?? 0),
-            workload: acc.workload + (c.average_workload ?? 0),
-            rating: acc.rating + (c.average_rating ?? 0),
-            skillsAreas: [...acc.skillsAreas, ...c.skills, ...c.areas],
-          },
+    (acc, c) => {
+      //const sectionIsInt = Number.isInteger(parseInt(c.section));
```
Josh-Cena, line 46: Don't leave commented out code in master branch

PR #1427, opened 2023-11-30
feat: URL social card images

bearsyankees:
Start to show course codes in URL previews


Josh-Cena:
Gonna close because this isn't the right direction for this particular goal

PR #1435, opened 2023-12-08
refactor: make sure we always capture Error instances

Josh-Cena:
Otherwise in Sentry, they always show up with type `<unknown>`.


bearsyankees (approved):
good fix

PR #1464, opened 2023-12-12
fix: make ICS export work again, merge logic for ICS/GCal

Josh-Cena:
Second part of https://github.com/coursetable/coursetable/issues/1448.

I made export ICS work again. I also made the two calendars share as much logic as possible. In addition I yanked the `ics` dependency because it doesn't do what I want and it's not hard to hand-roll ICS anyway.


frontend/src/components/Navbar/MeDropdown.tsx
```diff
@@ -49,36 +46,43 @@ function MeDropdown({
 }: Props) {
   // Fetch current device
   const { isMobile, isTablet } = useWindowDimensions();
+  const { cur_season, hidden_courses, courses } = useWorksheet();
```
bearsyankees, line 49: ya way better way to get season

frontend/src/config.ts
```diff
@@ -8,4 +8,116 @@ export const GRAPHQL_API_ENDPOINT = isDev
   ? 'https://localhost:8085'
   : import.meta.env.VITE_API_ENDPOINT + '/ferry';
 
+// Used for which season to show by default in catalog
```
bearsyankees, line 11: (and worksheet)

frontend/src/config.ts
```diff
@@ -8,4 +8,116 @@ export const GRAPHQL_API_ENDPOINT = isDev
...
+      {
+        name: 'Spring recess',
+        start: [2024, 3, 9],
+        end: [2024, 3, 25],
+      },
+    ],
+    transfers: [{ date: [2024, 1, 19], day: 1 }],
+  },
+  // Add more entries above, but don't remove any
+};
```
bearsyankees, line 123: This is all awesome. there also is another spot where we have years to see what seasons to fetch but i think its fine to leave that seperate

bearsyankees (approved):
this seems great and very helpful good stuff

PR #1468, opened 2023-12-17
refactor: move the last reference of year to config

Josh-Cena:
Now all year references live in the same file :)


bearsyankees:
🥚celent

PR #1472, opened 2023-12-20
downgrade passport version to 0.5.3

benzuzu:
we're getting this error https://github.com/jaredhanson/passport/issues/907, and it seems there is no valid fix out there atm. so we enforce the working version of passport 0.5.3 in the backend.


benzuzu:
https://github.com/coursetable/coursetable/pull/1471

fixed

Josh-Cena:
Honestly I should have gone for 0.5.3 instead of 0.5.0, but I guess it doesn't matter much now.

PR #1478, opened 2023-12-21
chore: fully onboard ESLint in frontend

Josh-Cena:
This is a one-off migration to strict ESLint settings! It's large and painful but it's done.


Josh-Cena:
It passed my cursory manual tests and I'm sure no one would bother to hand-review this; YOLO!

PR #1479, opened 2023-12-21
chore: fully onboard ESLint in api

Josh-Cena:
Like https://github.com/coursetable/coursetable/pull/1478 but for `/api`.


Josh-Cena:
YOLO

PR #1482, opened 2023-12-23
refactor: get rid of lodash in frontend

Josh-Cena:
This single change shaved 0.6MB from our bundle!

Before:

image

After:

image

I anticipate more to be shaved from lodash after we "fix" our other dependencies that transitively import lodash.

In this process I also fixed https://github.com/coursetable/coursetable/issues/1456 by using our own comparator instead of the default string comparison by `_.orderBy`.

image


Josh-Cena:
Celebrate!

image

PR #1484, opened 2023-12-23
feat: lazy-load route components

Josh-Cena:
Part of an effort to reduce bundle size. This PR makes all page components lazy-loaded, which means unless you are on the respective page, the respective code won't be loaded at all. This speeds up first page load speed, and completely avoids loading heavy dependencies like `graphiql` for 99% of our users.

Before:

image

After:

image

The total size is the same, but note how the main bundle is significantly smaller, and chunks can be loaded on-demand only.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -4,6 +4,10 @@ import { Routes, Route, Navigate, useLocation } from 'react-router-dom';
 import { Row, Spinner } from 'react-bootstrap';
 import * as Sentry from '@sentry/react';
 
+// Popular pages are eagerly fetched
+import Search from './pages/Search';
+import Worksheet from './pages/Worksheet';
+
 import Notice from './components/Notice';
 import Navbar from './components/Navbar/Navbar';
 import Footer from './components/Footer';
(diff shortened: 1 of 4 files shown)


bearsyankees:
is this that much faster? like at a glance it seems to load at basically the same speed for me as the normal CT but now it just takes a little more time to load the worksheet. But i get your logic here so maybe its worth it even if its just a tiny bit faster

Josh-Cena:
> it seems to load at basically the same speed for me as the normal CT

It's only noticeable on slower network indeed, but the point is we can now more easily optimize because there aren't as many critical dependencies, and subpages can use whatever heavy dependencies they want and those don't get loaded by default.

>  it just takes a little more time to load the worksheet

Yeah, I'll take some more time to check this. We need to preload the chunks so React never renders the loading screen at all. Not sure how to do that with suspense.

Josh-Cena:
@bearsyankees I fixed this by eagerly fetching Catalog/Worksheet instead.

bearsyankees:
ok ya this is way better. nice!!

Issue #1466, opened 2023-12-16
Get rid of luxon and moment


Josh-Cena:
We don't need third-party date libraries: they are heavyweight and contain lots of features we don't need (such as time zone conversion). We should be migrating to `Temporal` instead by installing a polyfill. (Or we can hand-roll our own date formatter instead, which shouldn't be too complicated.)

PR #1494, opened 2023-12-24
refactor: get rid of moment in frontend

Josh-Cena:
Yes, we did it!

Fixes https://github.com/coursetable/coursetable/issues/1466.

This even simplifies our code a bit, because we now use the `getCalendarEvents` utility.


Josh-Cena:
> -86.8 kB (-3%)

🎉

PR #1495, opened 2023-12-25
fix: friends security

bearsyankees:
Still needs some cleaning, let's not use get, and merge resolve friend and add friend, also can change params used now that we take netid from user but a start

### DIFF
--- a/api/src/friends/friends.handlers.ts
+++ b/api/src/friends/friends.handlers.ts
@@ -17,24 +17,20 @@ export const addFriend = async (
 
   const { netId } = req.user;
 
-  if (
-    !req.query ||
-    typeof netId !== 'string' ||
-    typeof req.query.id2 !== 'string'
-  )
+  if (!req.query || typeof req.query.id !== 'string')
     return res.status(400).json({ success: false });
 
   const friendNetId = req.query.id;
 
   // Make sure user has a friend request to accept
   try {
-    const friendRequest = await prisma.studentFriendRequests.findFirst({
+    const existingRequest = await prisma.studentFriendRequests.findFirst({
       where: {
         netId_friendNetId: { netId: friendNetId, friendNetId: netId },
       },
     });
 
-    if (!friendRequest) return res.status(400).json({ success: false });
+    if (!existingRequest) return res.status(400).json({ success: false });
   } catch (err) {
     winston.error(`Error with finding friend request: ${err}`);
     return res.status(500).json({ success: false });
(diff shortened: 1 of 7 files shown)


api/src/friends/friends.handlers.ts
```diff
@@ -13,19 +13,33 @@ export const addFriend = async (
 ): Promise<express.Response> => {
   winston.info('Adding new friend');
 
-  if (!req.user) return res.status(401).json({ success: false });
+  if (!req.user) return res.status(400).json({ success: false });
```
Josh-Cena, line 16:
```suggestion
  if (!req.user) return res.status(401).json({ error: 'USER_NOT_FOUND' });
```

api/src/friends/friends.handlers.ts
```diff
@@ -57,17 +71,17 @@ export const removeFriend = async (
 ): Promise<express.Response> => {
   winston.info('Removing friend');
 
-  if (!req.user) return res.status(401).json({ success: false });
+  if (!req.user) return res.status(400).json({ success: false });
```
Josh-Cena, line 74:
```suggestion
  if (!req.user) return res.status(401).json({ error: 'USER_NOT_FOUND' });
```

frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
```diff
@@ -343,6 +342,8 @@ export function NavbarWorksheetSearch() {
                         `Declined friend request: ${selectedOption.value}`,
                       );
                     }
+                    await resolveFriendRequest(selectedOption.value);
```
Josh-Cena, line 345:
```suggestion
                    // Resolve after add, so that when adding we can verify that request exists in DB
                    await resolveFriendRequest(selectedOption.value);
```

Josh-Cena (dismissed):
Make sure you run `bun install` because we just upgraded Prettier

api/src/friends/friends.handlers.ts
```diff
@@ -13,19 +13,33 @@ export const addFriend = async (
...
     typeof req.query.id2 !== 'string'
   )
-    return res.status(401).json({ success: false });
-
-  const netId = req.query.id2;
+    return res.status(400).json({ success: false });
 
   const friendNetId = req.query.id;
 
+  // make sure user has a friend request to accept
```
Josh-Cena, line 29:
```suggestion
  // Make sure user has a friend request to accept
```

Josh-Cena (approved):

PR #1508, opened 2023-12-25
feat: add a toast when failed to resolve worksheet course

Josh-Cena:
These errors aren't really actionable.

To test this, you can add a course manually:

`fetch("https://localhost:3001/api/user/toggleBookmark", { method: "POST", body: JSON.stringify({ action: "add", season: "202401", crn: 1, worksheetNumber: 0 }), headers: { "Content-Type": "application/json" }, credentials: "include" })`


Josh-Cena:
Actually we can't do this, because when rendering friends' worksheets, the same error will show up, but it will be unactionable. Maybe we should silence the error in that case.

Josh-Cena:
By running a quick script against the database, we figured that "dead courses" is actually a widespread issue and the average user doesn't really know how to act on it. Because we have got zero complaints of courses disappearing despite it constantly happening in the last 10 years, we think it's best to keep it this way and just silently remove nonexistent worksheet courses from our DB.

PR #1509, opened 2023-12-25
chore: remove unused data models

Josh-Cena:
I could be mistaken but these models seem to be unused. @bearsyankees @neilsong you can confirm


bearsyankees:
yeah i think ur right, from old/not implemented features

PR #1520, opened 2023-12-27
fix: browser storage may be inaccessible

Josh-Cena:
Session storage / Local storage is unavailable in Safari private mode as well as potentially other privacy modes.

In addition, I've delayed the retrieval of local storage to after page load. This is needed for server-side rendering.

Asking for feedback on whether the UX is good. Please add some filters and then reload the page.


bearsyankees:
What exactly are the UX changes here?

Josh-Cena:
@bearsyankees I thought there is, but looks like we show the loading screen while data is loading anyway (and display actual data way after first render), so there should be none.

Josh-Cena:
Works :) And since there's no UX degradation going to merge it as a simple bug fix

PR #1524, opened 2023-12-28
polish: add a counter for banner dismissal

Josh-Cena:


frontend/src/components/Notice.tsx
```diff
@@ -22,7 +23,13 @@ function Notice({ children }: { readonly children?: React.ReactNode }) {
       </div>
       {/* TODO */}
       {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
-      <span className={styles.closeButton} onClick={() => setVisible(false)}>
+      <span
+        className={styles.closeButton}
+        onClick={() => {
+          setVisible(false);
+          Sentry.captureMessage('Course evals banner dismissed');
```
bearsyankees, line 30: could have this be generic and forever but ya seems good

bearsyankees (approved):

frontend/src/components/Notice.tsx
Josh-Cena: Nah, we want to recollect analytics for every banner, no?

PR #1525, opened 2023-12-28
refactor: use banner storage strategy that's not linear-growing

Josh-Cena:
This uses a single storage key instead of requesting a new storage key each time, which means we don't have to clear previous slots.


bearsyankees:
:)

PR #1536, opened 2023-12-30
feat: SSR POC

Josh-Cena:
This is needed for https://github.com/coursetable/coursetable/pull/1427 to work.

@bearsyankees My website is fully server-rendered with react-helmet. You can check the impl here: 

- https://github.com/Josh-Cena/Josh-Cena.github.io/blob/master/server/pre-render.ts
- https://github.com/Josh-Cena/Josh-Cena.github.io/blob/master/src/server-entry.tsx

The code here is slightly adapted to make bun work.

Next we need to figure out how to serve these pre-rendered HTML so they get picked up by social media.


Josh-Cena:
I haven't wired it to any deployment yet. To test the build output, run `cd frontend && bun run build:ssr && bun src/ssr/pre-render.ts`.

bearsyankees:
@Josh-Cena is this still viable/ something we want to do

Josh-Cena:
I definitely want to do this eventually, but not sure if it yields any immediate value. We have to migrate off Vercel static serving first.

PR #1539, opened 2023-12-31
docs: add documentation for container/DB setup

Josh-Cena:
@bearsyankees @neilsong Maybe you want to proofread it, and/or you have anything you want to add?


neilsong:
Regarding the course data workflow:

What I'm trying to do with the ferry migration is get rid of the DB stuff completely. As you pointed out in the `TODO` we should get rid of Postgres - 100% with that. My idea with the course data is keep the JSON, since the logic for generating that nice and stable. Cut the GraphQL + Postgres middleman and statically serve the JSONs behind a blob storage endpoint that will be abstracted away in one GET request with a unique URL as a Doppler secret. This way we can thin out DB dependencies and reuse all the JSON logic that exists. What do you think?

If we really wanna drill down on security, it is also possible to restrict access to this URL to only the production containers and devs with a VPN connection into the VNet. This seems overkill for dev though.

docs/containers.md
```diff
@@ -0,0 +1,39 @@
...
+1. Ferry writes course data to [`ferry-data`](https://github.com/coursetable/ferry-data) in JSON.
+2. Ferry then writes the data to a Postgres database on the prod server.
+3. The Postgres database exposes itself to the Internet.
+4. During development and prod, we spin up a Hasura engine that wraps the Postgres database and exposes a GraphQL API.
+5. The Express app queries the GraphQL API and generates static JSON again, located in the `api/static` folder. These are cached locally unless you run `./start.sh -d -o`.
+6. The frontend requests the Express endpoint, which serves these JSON files.
+
+And for user data:
+
+1. In dev, we always create a new MySQL database; in prod, we use the existing one. The table schema of both should be consistent with `api/mysql/database.sql`.
```
Josh-Cena, line 25: @neilsong The dev DB \*uses\* `api/mysql/database.sql`. Maybe you mean `api/prisma/schema.prisma`? (All three should be synced anyway)

Josh-Cena:
> Cut the GraphQL + Postgres middleman

Strong -1. I see great benefits in GraphQL + DB. Several things:

1. We can provide "progressive enhancement" where you can view publicly-accessible data such as course title & description even without login.
2. We can query the DB if we want to analyze something ourselves, such as figuring out what a particular CRN maps to.
3. We have open issues for exposing the GraphQL endpoint and I'm sure this can easily benefit others for their research.

docs/containers.md
neilsong: Ah yea, table schema here is kind of an ambiguous term. The DB's true table schema is defined through executing database.sql but is then described in prisma language in schema.prisma. Either way, I agree that all three must be consistent - feel free to change that line however you see fit.

neilsong:
I see - if we integrate the DB with our MySQL setup, we should be ok. There's some noodling around I need to do with the ferry refresh script in a cloud env to make it work but should be good.

If I'm getting this right, the idea is to use the same MySQL DB behind 2 methods of access:

1. Express with a Prisma MySQL DB connector (essentially a direction connection)
2. A public GraphQL endpoint for course data retrieval (that express also uses instead of the direct connection?)

We definitely need to restrict the scope of that GraphQL endpoint but LGTM otherwise.

Josh-Cena:
Yep, exactly my idea.

neilsong:
So for the course data retrieval in express, we just do it on express startup and keep it memory I assume.

neilsong:
@bearsyankees Looks like a solid plan to me - I can also use this opportunity to rewrite ferry logic to be concurrent + stateless.

Goal would be to make ferry an ephemeral container that serves the purpose of a script. Azure Functions should handle the daily/weekly deployments.

bearsyankees:
Ok this seems good. When talking about the "public GraphQL endpoint", public to yale students (so they can get the review data still), or to general public just for seeing names of courses and stuff, or both? Other than that this seems good, i am pro getting rid of the json (and takes up space), as long as we still have a good way to share data w folk who can use it. And this will make it easier to manually link courses that should be linked. But this may be a big undertaking to rewrite ferry.....

Epic team thank u guys @neilsong @Josh-Cena  fr

neilsong:
I don't have a particular opinion regarding the public endpoint.

As for JSON backwards compatibility, we can always just upload the current catalog files to a static storage point and keep that link.

Josh-Cena:
I really hope that the frontend can directly query the course data with GraphQL instead of as static JSON. We should only do the latter if GraphQL responses cannot be cached (which I've heard is the case).

PR #1546, opened 2024-01-01
feat: make everything use 🍞

Josh-Cena:
All scripts now use Bun.

cc @neilsong


api/package.json
neilsong, line 1: We should be all good if all the node-native API's we use have full coverage/implementation in Bun. I think I checked this at some point, so we should be ok

neilsong, line 1: We should be all good if all the node-native API's we use have full coverage/implementation in Bun. I think I checked this at some point, so we should be ok

api/Dockerfile
neilsong, line 1: I think we can cache better with multi-stage builds, I added a commit.

@Josh-Cena Can you run this commit on your computer? My bun install is segfaulting in the Docker build but apparently it might be because the public wifi I'm on doesn't support ipv6.

api/package.json
Josh-Cena: Yes, I think it's the case. Only very obscene V8 APIs don't have implementation.

api/Dockerfile
Josh-Cena: Unfortunately China blocked access to githubusercontent.com, and I don't know how to forward my network proxy to my docker... My Linux remote segfaults on bun install too, but it comes with feeble RAM so probably not bun's fault either.

neilsong: Can confirm bun install does not segfault.

neilsong:
For some reason, api auth hangs on "Logging in with CAS." Not quite sure why - nothing else is logged.

api/Dockerfile
```diff
@@ -1,33 +1,29 @@
 # Remember: the context for this build is the root of the monorepo,
 # not the api folder
 
-# TODO: we only use bun as a dependency manager, but api is still executed with
-# node. We should eventually get rid of node
-FROM oven/bun:slim as base
+# We need Node to run prisma generate. See https://github.com/oven-sh/bun/issues/3827
+# TODO: switch to oven/bun-node:slim
+FROM imbios/bun-node:latest-current-slim as base
```
Josh-Cena, line 6: @neilsong: I purposefully used `oven/bun` directly so there's minimal underlying magic and future migration less break-prone. Does nvm not work?

neilsong: Yea for some reason it’s hella cooked for me. Prisma is being cringe as usual.

Future migration should be a simple change from this image to bun. This image only just has node and bun side-by-side.

neilsong: If you want, we can take a look at that image and copy over their code for installing node with bun but I feel like that’s kinda unnecessarily reinventing the wheel ykwim?

Josh-Cena: I know, I started by using this image. But:

1. No control of Node/Bun version and no guarantee that it will be latest and/or customizable
2. It's 3rd party

neilsong: Alri - I mean it uses a GH action to automatically tag and push versioned images based on bun updates but it could break if bun changes their update process.

I’ll take a look at their Dockerfile and see what I can do

neilsong: Their dockerfile installs bun on top of node.

Since we only need the node binary, let’s just copy it over from the node docker image into the base image - what do you think?

Josh-Cena: Erm—I still felt like using nvm is the most controllable and robust way, but otherwise using this image is probably better.

api/Dockerfile
```diff
@@ -1,33 +1,29 @@
 # Remember: the context for this build is the root of the monorepo,
 # not the api folder
 
-# TODO: we only use bun as a dependency manager, but api is still executed with
-# node. We should eventually get rid of node
-FROM oven/bun:slim as base
+# We need Node to run prisma generate. See https://github.com/oven-sh/bun/issues/3827
+# TODO: switch to oven/bun-node:slim
```
Josh-Cena, line 5: In any case shouldn't our eventual goal be to use `oven/bun:slim` and remove all dependency on Node?

neilsong: Oh that's my bad, I meant to write bun not bun-node

api/Dockerfile
neilsong: kk - i'll work out nvm

neilsong: ok this thing is getting way too complicated because of the way docker manages shells and sessions. nvm relies on session-specific env vars that get wiped in weird ways cuz docker. i think we just go with this image.

api/Dockerfile
neilsong, line 1: @Josh-Cena What about this approach?

api/Dockerfile
```diff
@@ -1,33 +1,35 @@
...
 RUN cd /temp/dev && bun install --frozen-lockfile
 
-# Copy dependencies and source code into final image
-FROM runtime AS final
+# We need Node to run prisma generate. See https://github.com/oven-sh/bun/issues/3827
+# TODO: delete this
+FROM node:slim as prisma
+COPY --from=install /temp/dev/node_modules /temp/dev/node_modules
+COPY api/prisma/ /temp/dev/prisma/
+RUN cd /temp/dev && npx prisma generate
```
neilsong, line 22: @Josh-Cena What about this approach?

Josh-Cena: LGTM

Josh-Cena (commented):
I'll be barely reachable in the next few days. Marking my approval and you may merge

neilsong:
ty for your help 🫡

neilsong:
> For some reason, api auth hangs on "Logging in with CAS." Not quite sure why - nothing else is logged.

Remains an issue for me, would be great if someone could double-check.

bearsyankees:
is this prod ready ish? or where do we stand @neilsong @Josh-Cena

Josh-Cena:
I just need others to confirm that it works on their machines too.

neilsong:
Login does not redirect successfully to catalog - goes back to home page.

> express         | {"level":"error","message":"Yalies connection error: Error: Forbidden","timestamp":"2024-01-28 21:38:37:3837"}

Josh-Cena:
Oh no, it's back again...

frontend/tsconfig.json
```diff
@@ -7,8 +7,8 @@
     "moduleResolution": "bundler",
     // We use ESLint for this
     "allowUnreachableCode": true,
-    "jsx": "react",
+    "jsx": "react"
```
Josh-Cena, line 10: @neilsong Something is wrong here. These commas should be there!

frontend/src/utilities/course.ts
```diff
@@ -138,7 +138,7 @@ export function getNumFriends(friends: FriendRecord): NumFriendsReturn {
           const key = seasonCode + course.crn; // Key of object is season code + crn
           // There are a lot of ESLint bugs with index signatures and
           // no-unnecessary-condition
-          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+
```
Josh-Cena, line 141: I don't think this lint error should be gone; if it is indeed, then please remove the comments above too

neilsong: Will resolve this on master 👍

frontend/tsconfig.json
neilsong: Check master - it's a prettier update and I just did a clean install today and was really confused by this. I'm going to reconcile these branches in a second.

neilsong:
Now passing login (#1595)

Fails on opening course modal:
```
express         | 49 |                 this.catchUpgradeRequest(server);
express         | 50 |             }
express         | 51 |         };
express         | 52 |         this.catchUpgradeRequest = (server) => {
express         | 53 |             if (!this.wsInternalSubscribed) {
express         | 54 |                 server.on('upgrade', this.handleUpgrade);
express         |                      ^
express         | TypeError: undefined is not an object (evaluating 'server.on')
express         |       at /usr/src/app/node_modules/http-proxy-middleware/dist/http-proxy-middleware.js:54:17
express         |       at /usr/src/app/node_modules/http-proxy-middleware/dist/http-proxy-middleware.js:49:17
```

Josh-Cena:
Fun! LMK if you need help. At least an error is easier to debug than nothing showing up

neilsong:
Dug a little deeper - for some reason, this ternary doesn't get the internal express server object properly

```const server = (_b = ((_a = req.socket) !== null && _a !== void 0 ? _a : req.connection)) === null || _b === void 0 ? void 0 : _b.server;

neilsong:
Build size report is failing because it attempts to run the `patch-package` postinstall step in base when building base for comparison. As soon as base is updated, the check will pass.

PR #1547, opened 2024-01-02
chore: add diff check to frontend cd and deploy check to preview comment

neilsong:


neilsong:
rip no more wifi - only thing left is to add the diff check to staging and prod cd

Josh-Cena:
@neilsong Perhaps you should use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore instead?

neilsong:
ty @Josh-Cena - will take a look

Issue #1551, opened 2024-01-15
GCal `colorId` cannot increase infinitely


Josh-Cena:
## Bug description

When exporting to gcal, we assign a unique `colorId` to each event. However, there are only finitely many `colorId` available for each calendar.

## Reproduction

1. Add >15 class to your calendar
2. Export to GCal
3. Sentry captures the following error:

  ```json
  {
    "error": {
      "errors": [
        {
          "domain": "global",
          "reason": "invalid",
          "message": "Invalid Value"
        }
      ],
      "code": 400,
      "message": "Invalid Value"
    }
  }
  ```

See: https://coursetable.sentry.io/issues/4865521285/?project=5515218&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=23

We have some solutions here:

1. Use https://developers.google.com/calendar/api/v3/reference/colors/get to first fetch all colors available
2. Use the same color for all events

PR #1552, opened 2024-01-15
fix: cycle gcal event colorId

benzuzu:
the colorid will no longer go to infinity, so it will always be valid.

### DIFF
--- a/frontend/src/utilities/calendar.ts
+++ b/frontend/src/utilities/calendar.ts
@@ -139,7 +139,7 @@ function toGCalEvent({
       timeZone: 'America/New_York',
     },
     recurrence,
-    colorId: (colorIndex + 1).toString(),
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
FYI you can use "Fixes #1551" in your issue description which will autoclose the issue when the PR is merged

frontend/src/utilities/calendar.ts
```diff
@@ -139,7 +139,7 @@ function toGCalEvent({
       timeZone: 'America/New_York',
     },
     recurrence,
-    colorId: (colorIndex + 1).toString(),
+    colorId: ((colorIndex % 11) + 1).toString(),
```
Josh-Cena, line 142:
```suggestion
    // Cycle the ID because each calendar only has a fixed set of colors
    colorId: ((colorIndex % 11) + 1).toString(),
```

PR #1558, opened 2024-01-17
fix: GraphiQL use default CSS

Josh-Cena:
I don't know why we are self-hosting a stylesheet but using the default feels so much more correct...

Check out https://fix-graphiql.preview.coursetable.com/graphiql


Josh-Cena:
The doubled CSS bundle is unfortunate, though :( I will soon look into CSS code splitting!

bearsyankees:
lets go this looks great i was always frustrated by that interface @Josh-Cena

PR #1581, opened 2024-01-30
fix: consistently use the currently viewed worksheet to calculate conflicts

Josh-Cena:
Currently we always use the main worksheet, which looks non-ideal because I may be bluebooking using another worksheet.


bearsyankees:
LGTM, just wondering if we should have some indicator or way to change what worksheet ur on on the catalog page too so they don't need to switch to worksheet first to change, but could just get clunky...

Josh-Cena:
Yeah, I think we should eventually do that.

Issue #1527, opened 2023-12-29
Store worksheet course hidden status & color in DB


Josh-Cena:
This should allow us to fix https://github.com/coursetable/coursetable/issues/1441 & https://github.com/coursetable/coursetable/issues/1443

bearsyankees:
@Josh-Cena do we want to also still use local storage for hidden courses, or just use completely rely on db for it?

Josh-Cena:
Let's completely rely on DB. You have to use the DB to get the list anyway.

Josh-Cena:
https://github.com/coursetable/coursetable/pull/1661 starts the effort

Josh-Cena:
Going to close this and track the rest in https://github.com/coursetable/coursetable/issues/1441

PR #1583, opened 2024-01-30
feat: fetch worksheet color from DB

Josh-Cena:
Part of https://github.com/coursetable/coursetable/issues/1527


api/mysql/database.sql
```diff
@@ -73,7 +73,8 @@ CREATE TABLE `WorksheetCourses` (
   `netId` char(8) NOT NULL,
   `crn` mediumint(8) UNSIGNED NOT NULL,
   `season` mediumint(8) UNSIGNED NOT NULL,
-  `worksheetNumber` mediumint(8) UNSIGNED DEFAULT 0
+  `worksheetNumber` mediumint(8) UNSIGNED DEFAULT 0 COMMENT 'Multiple worksheets!',
```
bearsyankees, line 76:
```suggestion
  `worksheetNumber` mediumint(8) UNSIGNED DEFAULT 0 COMMENT 'Indicator of which of the user's worksheets this belongs to',
```

api/prisma/schema.prisma
```diff
@@ -46,6 +46,7 @@ model WorksheetCourses {
   crn             Int    @db.UnsignedMediumInt
   season          Int    @db.UnsignedMediumInt
   worksheetNumber Int    @default(0) @db.UnsignedMediumInt
+  color           String @db.VarChar(32)
```
bearsyankees, line 49: using string should probably be fine, but there is some native hex support (the benefits of which i haven't totally looked into)
https://stackoverflow.com/questions/4306219/inserting-hex-value-in-mysql

api/src/user/user.handlers.ts
```diff
@@ -121,6 +122,6 @@ export const getUserWorksheet = async (
     evaluationsEnabled: studentProfile?.evaluationsEnabled ?? null,
     year: studentProfile?.year ?? null,
     school: studentProfile?.school ?? null,
-    data: worksheetCoursesToWorksheets(worksheets)[netId]!,
+    data: worksheetCoursesToWorksheets(worksheets)[netId] ?? {},
```
bearsyankees, line 125: yay good safety

api/prisma/schema.prisma
bearsyankees: (if we ever allow users total control over the color, this could be helpful to make sure we aren't storing non hex values in this column)

bearsyankees (changes requested):
this looks solid nothing major just left a few comments. great work standardizing the Listing too, should make further changes much easier

bearsyankees:
although we get an error on worksheet on ur preview but i assume bc its trying to access columns of the db that don't exist

bearsyankees:
image
this is whats logged from the preview error

api/mysql/database.sql
Josh-Cena: @bearsyankees This is the same comment as in the prod DB :P

Josh-Cena:
Yes, because the backend isn't deployed. You can only preview this locally.

api/mysql/database.sql
bearsyankees: ya and i made the prod one but i think this is more descriptive lol i was lazy

api/prisma/schema.prisma
Josh-Cena: Nah, we aren't _just_ going to use hex. If user specified a `hsl` color we are going to use that.

Josh-Cena: It's not going to be lossless so I'd rather not. I'm confident that precisely zero person is going to care but just to be safe.

bearsyankees: kk sounds good.  we should do some backend validation too though that what is storing is valid

Josh-Cena: Good idea

bearsyankees:
aight deploy after our meeting today and make sure backend is good?

Josh-Cena:
Sure!

PR #1586, opened 2024-02-01
feat: search location

Josh-Cena:
Part of https://github.com/coursetable/coursetable/issues/1453

### DIFF
--- a/frontend/src/contexts/searchContext.tsx
+++ b/frontend/src/contexts/searchContext.tsx
@@ -513,7 +513,23 @@ export function SearchProvider({
           listing.title.toLowerCase().includes(token) ||
           listing.professor_names.some((professor) =>
             professor.toLowerCase().includes(token),
-          )
+          ) ||
+          // Use `times_by_day` instead of `locations_summary` to account for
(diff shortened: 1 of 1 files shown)


bearsyankees:
LGTM but do we want a toggle for this or just alwayson?

Josh-Cena:
I implemented this in a way that shouldn't cause any confusion. The building names almost always never collides with other useful information, so we can just do this.

bearsyankees:
Ok yeah. How should we alert people that this is an option now though

Josh-Cena:
> How should we alert people that this is an option now though

I mean... The text box says "or whatever we don't care" so they know when they try?

bearsyankees:
I guess but I feel like no one will wake up and try by location.... but I don't see an amazing solution

Josh-Cena:
I mean... I've tried, because I thought all displayed text is searchable. I guess that's a reasonable expectation. Few people probably need this anyway. (I needed this because I often wonder who is having class in our classroom before and after a class.)

bearsyankees:
OKEDOKE i give in

PR #1589, opened 2024-02-01
refactor: use native history stack for back operation instead of custom code

Josh-Cena:


Josh-Cena:
I decided that the current way is better.

PR #1595, opened 2024-02-06
refactor: use express-session instead of cookie-session and cookie-parser

neilsong:
`express-session` is a superset of the functionality of `cookie-session`. It removes the dependency on `cookie-parser` and is required to solve the login issue in #1546


api/src/server.ts
```diff
@@ -62,19 +61,25 @@ app.use((req, _, next) => {
 
 // Setup sessions.
 app.use(
-  cookieSession({
+  session({
```
neilsong, line 64: Need some help on solving the eslint error here :)

Josh-Cena:
@neilsong FYI: whenever you add a new module, make sure you have its `@types` (unless the package provides native types). If you remove a module, remove its `@types` too. The `unsafe-call` is because that `session` function has no types so it becomes `any`.

neilsong:
Gotcha, thank you so much!

Josh-Cena:
Could you tell me a bit more about what doesn't work with `cookie-parser`?

neilsong:
express-session just takes care of cookie-parser, I assume they just implemented all the functionality.

Josh-Cena:
Have you tested it? If so I'm not going to duplicate the effort.

neilsong:
I've tested it - basic auth functionality, loading courses, and adding them works. Outside of that, I'm not sure.

Josh-Cena (approved):

PR #1604, opened 2024-02-13
feat: redesign modal, make evaluations view easier to find

Josh-Cena:
Frees up a lot of horizontal space for the title row.

Also, we heard complaints that not many read the evaluations. We hope that making it a top-level selection is going to improve the situation.

Before:

image

After:

image

### DIFF
--- a/frontend/src/components/Search/CourseConflictIcon.tsx
+++ b/frontend/src/components/Search/CourseConflictIcon.tsx
@@ -11,7 +11,7 @@ import {
   checkCrossListed,
   isInWorksheet,
 } from '../../utilities/course';
-import { CUR_SEASON } from '../../config';
+import { CUR_YEAR } from '../../config';
 
 interface CourseConflictIconProps {
   readonly course: Listing;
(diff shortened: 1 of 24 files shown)


bearsyankees:
Not seeing any evals on preview branch, in class now tho will try and debug after

Josh-Cena:
@bearsyankees You are viewing spring 2024, which has no evals :)

bearsyankees:
Ah so you would just show evals for that year? I don't think students go back to look for evals like that, we should display most recent first I guess and have a drop down to shift, but we already kinda have that on the side............ i will ponder the UI/UX for this after class. But i dont think its that useful to have to go back to 2023 to see the evals. (and havent looked at the code yet so my bad for misunderstanding the desired effect)

Josh-Cena:
@bearsyankees The idea is that every course has one associated "overview" and "evaluations" page. You don't just randomly draw evaluations from another course that happens to have the same course code.

bearsyankees:
Let's discuss face2face and maybe get some thoughts from the team

frontend/src/components/Search/CourseConflictIcon.tsx
```diff
@@ -47,26 +56,28 @@ function CourseConflictIcon({ course }: { readonly course: Listing }) {
     [course, data],
   );
 
+  const seasonMismatch = course.season_code !== CUR_SEASON;
```
Josh-Cena, line 59: We can't do this because students in the fall season may still want to plan some spring classes; we should use `CUR_YEAR.includes(course.season_code)`

frontend/src/components/Search/CourseConflictIcon.tsx
```diff
@@ -47,26 +56,28 @@ function CourseConflictIcon({ course }: { readonly course: Listing }) {
...
+        {displayConflict && (
           <OverlayTrigger
             placement="top"
             overlay={(props) => (
               <Tooltip {...props} id="conflict-icon-button-tooltip">
                 <small style={{ fontWeight: 500 }}>
-                  Conflicts with: <br />
-                  {conflicts.map((x) => x.course_code).join(', ')} <br />
+                  {seasonMismatch
+                    ? `This will add this course to a worksheet in a different season.`
```
Josh-Cena, line 75: Not "in a different season"; we always add to the season the course is in. It's just not the "current season".

bearsyankees (approved):

neilsong:
Do you think we could lessen the gap/padding on the right of the share icon? Looks a bit odd since there's not much padding on the left side of the tabs. It's a pretty minor thing - just follows the left/right alignment pattern we got going on. I'm not strongly opinionated either way though.

Josh-Cena:
No strong feelings for me either. I think it just because of the close button taking up width.

PR #1612, opened 2024-02-19
feat: friend request notif icon

bearsyankees:
Should be good just need some UI tips on where exactly it should go...


bearsyankees:
eh should do numbers and not red maybe

Josh-Cena (approved):

PR #1617, opened 2024-02-26
feat: public courses

### LINKED ISSUE #1616: Make Public Course Info Public
If no auth on login, show course data just no evals like YCS does

https://coursetable.canny.io/admin/board/feature-requests/p/make-public-course-info-public

bearsyankees:
work in progress fixes #1616


bearsyankees:
bruh idk why has the quist stuff but whatever its up to date

api/src/server.ts
```diff
@@ -148,6 +148,17 @@ friends(app);
 canny(app);
 user(app);
 
+// Serve public catalog files without authentication
+app.use(
+  '/api/static/catalogs/public',
```
Josh-Cena, line 153: @bearsyankees I think the issue is that the `/api/static/catalogs` is a parent path of this, so if the former requires auth, this will too. I suggest moving this to something like `/api/static-public`, or you would have to move the currently needing-auth data to a subdirectory.

Also, I suggest reusing the data as much as possible, and only merge the extra data, instead of using an entirely different set of data.

bearsyankees: No this works for auth. So you don't think having just 2 jsons is the way to go? don't think it takes up that much space but i hear u def could be better to always pull from public except for eval data... imma try and get this working first then figure out how to optimize

api/src/catalog/catalog.utils.ts
```diff
@@ -92,39 +96,68 @@ export async function fetchCatalog(
...
+    const publicCatalogPath = `${STATIC_FILE_DIR}/catalogs/public/${seasonCode}.json`;
+    let catalogFull: Catalog = { computed_listing_info: [] };
+    let catalogPublic: Catalog = { computed_listing_info: [] };
 
+    // Fetch and save the full catalog (including evaluations)
     if (
-      !overwrite &&
-      (await fs.access(outputPath).then(
+      overwrite ||
+      !(await fs.access(fullCatalogPath).then(
```
bearsyankees, line 107: maybe this should go back to 
`!overwrite &&
      (await fs.access(outputPath).then(`

bearsyankees:
Still need to render the modal how we want it, and show the filter bar at the top and have no error when you try and add the course, maybe we need a totally new public result page and public modal component

bearsyankees:
need to configure hasura re graphql auth, shouldn't be bad

api/src/server.ts
```diff
@@ -115,22 +115,25 @@
 passportConfig(passport);
 app.use(passport.initialize());
 app.use(passport.authenticate('session'));
+app.use((req, res, next) => {
+  req.headers['X-Hasura-Role'] = req.isAuthenticated()
+    ? 'student'
+    : 'anonymous';
+  console.log('Forwarding role:', req.headers['X-Hasura-Role']);
+  next();
+});
```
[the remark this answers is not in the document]
bearsyankees: TRUE!!!

bearsyankees:
now also fixes #1622 kinda

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -98,6 +100,17 @@ const CourseModalEvaluations = suspended(
 function CourseModal() {
   const [searchParams, setSearchParams] = useSearchParams();
   const { requestSeasons, courses } = useFerry();
+  const [isAuthenticated, setIsAuthenticated] = useState<boolean | undefined>(
+    undefined,
+  );
+  useEffect(() => {
+    // Better way to do auth?
```
Josh-Cena, line 107: We should set this as a context value.

PR #1624, opened 2024-03-04
feat: worksheet split view on catalog page

BradleyLewis08:
## Pre-flight checklist

- [ ] I have searched for potential duplicate pull requests.
- [ ] **If this is a code change**: I have written unit tests and/or conducted manual tests to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes https://github.com/coursetable/coursetable/issues/1610) and the maintainers have approved on my working plan.

## Summary

This PR fixes/implements the following feature:

- Worksheet Split View


bearsyankees:
image
Can we make the worksheet preview bigger so i can see the full thing? this looks epic so far great stuff

bearsyankees (changes requested):
(see above)

frontend/src/pages/OverlayComponent.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useEffect } from 'react';
```
Josh-Cena, line 1: Move these files to `components`; `pages` are for the route components

Josh-Cena:
Hi @BradleyLewis08 @bearsyankees I've made some edits.

- Most notably: I'm now using `WorksheetCalendar` instead of `Worksheet`. This means we don't need to hook into `worksheetView`.
  - Why? Because otherwise when you navigate away from this page, the worksheet view stays as `overlay`, and we don't know which view to restore it too!
  - This is up for discussion: will people ever want to see the worksheet in "list" view?
  - Will they want to see a list of courses they've added to the worksheet, like the right sidebar on the worksheet page?
  - My answer is "no", and it's too hard to design it otherwise.
- I've added mobile support (and support in grid view).
- The calendar bottom no longer overflows the screen.
- Little CSS improvements.
- Moved some files

LMK if they look good

Josh-Cena:
For a stable branch preview, please use the following URL: https://ewis-worksheet-split-view.preview.coursetable.com/catalog

Josh-Cena (approved):
Rubber stamping: effect looks good to me now

neilsong:
Two things before it LGTM:

1. Let's make the fullscreen icon padded from the top border
image

2. Let's persist the state of the split view (shown or hidden) after navigating to and from a different page

Other than that - good stuff looks awesome

Josh-Cena:
@neilsong Please take another look

neilsong:
LGTM thank you for your commits :pray:

neilsong (approved):

bearsyankees:
@neilsong @Josh-Cena only 1 thing from me -- do we still want it on public view (as it is right now)? I am not entirely opposed as it shows what features they could have if they sign in, but also is not useful at all as of now without public worksheet

Josh-Cena:
I think this one is fine. The idea is to eventually have as little branching logic as possible except for things unauthorized users cannot see.

bearsyankees:
> I think this one is fine. The idea is to eventually have as little branching logic as possible except for things unauthorized users cannot see.

OK LGTM then

bearsyankees (approved):

PR #1627, opened 2024-03-05
feat: gcal enhancement

benzuzu:
A few enhancements to Gcal integration:

- transfer days (friday becoming monday) figured out
- spring 2025 calendar added
- due to popular demand, we add more details to the gcal events, specifically course + instructor name.
Screenshot 2024-03-20 at 10 08 10 PM

### DIFF
--- a/frontend/src/config.ts
+++ b/frontend/src/config.ts
@@ -51,17 +51,17 @@ export const academicCalendars: { [season: Season]: SeasonCalendar } = {
       {
         name: 'Labor Day',
         start: [2022, 9, 5],
-        end: [2022, 9, 6],
+        end: [2022, 9, 5],
       },
       {
         name: 'October recess',
         start: [2022, 10, 19],
-        end: [2022, 10, 24],
+        end: [2022, 10, 23],
       },
       {
         name: 'November recess',
         start: [2022, 11, 19],
-        end: [2022, 11, 28],
+        end: [2022, 11, 27],
       },
     ],
     transfers: [{ date: [2022, 9, 2], day: 1 }],
@@ -73,12 +73,12 @@ export const academicCalendars: { [season: Season]: SeasonCalendar } = {
       {
         name: 'Martin Luther King Jr. Day',
         start: [2023, 1, 16],
-        end: [2023, 1, 17],
+        end: [2023, 1, 16],
       },
       {
         name: 'Spring recess',
         start: [2023, 3, 11],
-        end: [2023, 3, 27],
+        end: [2023, 3, 26],
       },
     ],
     transfers: [{ date: [2023, 1, 20], day: 1 }],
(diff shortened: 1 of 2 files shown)


frontend/src/utilities/calendar.ts
```diff
@@ -281,7 +281,7 @@ export function getCalendarEvents(
             `RRULE:FREQ=WEEKLY;BYDAY=${byDay};UNTIL=${endRepeat}Z`,
             `EXDATE;TZID=America/New_York:${exDate}`,
           ],
-          description: c.title,
+          description: `Course: ${c.title}\nInstructor: ${c.professor_names.join(', ')}`,
```
Josh-Cena, line 284:
```suggestion
          description: `${c.title}\nBy ${c.professor_names.join(', ')}`,
```

What do you think?

benzuzu: Considered doing this or "with <instructor>, but there are some edge cases where merging the two would sound weird (i.e. if the course name has "by", "with", etc.)

Josh-Cena: Yeah, but the line break would fix the issue with ambiguity, I hope. Otherwise "Instructors" is also fine, I just want to get rid of the "Course" part

frontend/src/utilities/calendar.ts
```diff
@@ -100,14 +100,26 @@ function datesInBreak(
...
     return dates;
   });
 }
 
+function transferDays(
+  transfers: SeasonCalendar['transfers'],
+  days: number[],
+  time: string,
+) {
+  return transfers.flatMap((t) => {
```
Josh-Cena, line 116: You are only returning strings. Shouldn't this be `map` instead of `flatMap`?

benzuzu: right, either works in this case but `map` makes more sense.

frontend/src/config.ts
```diff
@@ -129,21 +129,38 @@ export const academicCalendars: { [season: Season]: SeasonCalendar } = {
...
+        name: 'Spring recess',
+        start: [2025, 3, 8],
+        end: [2025, 3, 23],
+      },
+    ],
+    transfers: [{ date: [2025, 1, 24], day: 1 }],
+  },
   // Add more entries above, but don't remove any
-  // Use zero-based indexing, i.e. "start" included, "end" excluded
+  // Use fully-closed indexing, i.e. "start" and "end" both included
```
Josh-Cena, line 165: @benzuzu This is perhaps slightly confusing but the academic calendar itself is inconsistent. The dates for each break are left-right-exclusive ("Oct 17 – October recess begins, after last academic obligation"; "Oct 23 – classes resume"), while semesters are left-right-inclusive ("Apr 26 – classes end, 5:30pm"). In an effort to uniformize, I picked left-inclusive right-exclusive because that's the most typical boundary convention in JavaScript and other C-like languages. One has to do translations anyway even if both boundaries are inclusive.

Josh-Cena (changes requested):

frontend/src/config.ts
benzuzu: @Josh-Cena Are they not both left-right-inclusive now? I believe I changed it, otherwise labor day would not be working. 

`name: 'Labor Day',
        start: [2024, 9, 2],
        end: [2024, 9, 2],`

Josh-Cena: Hmm, sure, no strong opinions. I'll soon refactor this to be dynamic instead of hardcoded anyway.

Josh-Cena (approved):
Thanks! LGTM

Josh-Cena:
@benzuzu I realized this would affect the worksheet calendar too. Was that your intention? It looks slightly confusing there IMO. If not, I can quickly make a fix for the worksheet calendar.

benzuzu:
> @benzuzu I realized this would affect the worksheet calendar too. Was that your intention? It looks slightly confusing there IMO. If not, I can quickly make a fix for the worksheet calendar.

ah yes, definitely not. sorry, didn't realize that was also changed in your refactor earlier. fix would be appreciated.

PR #1629, opened 2024-03-07
polish: better public UI

bearsyankees:
better ui


Redorhcs:
background: original random colors could be perceived as actual ratings. Shifted color scheme to ct blue/grey. looks great!

bearsyankees:
> background: original random colors could be perceived as actual ratings. Shifted color scheme to ct blue/grey. looks great!

Thank you for your help tyler :))

PR #1632, opened 2024-03-11
feat: show percentage of enrolled students who left review

Josh-Cena:
Fix https://github.com/coursetable/coursetable/issues/1620

### DIFF
--- a/frontend/src/components/CourseModal/EvaluationRatings.tsx
+++ b/frontend/src/components/CourseModal/EvaluationRatings.tsx
@@ -58,6 +58,7 @@ function EvaluationRatings({
           ratings={filteredRatings[question]}
           reverse={question === 'workload' || question === 'challenge'}
           labels={evalQuestions[question].labels}
+          enrolled={info.enrolled ?? 0}
         />
       </div>
     ));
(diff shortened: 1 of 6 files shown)


bearsyankees:
Looks good but i think we should make it even more prominent, and maybe somehow show it also on the catalog page (maybe as a tooltip hover on the overall/work/prof)?

Josh-Cena:
@bearsyankees The catalog view shows average across seasons, not this particular offering.

bearsyankees:
> @bearsyankees The catalog view shows average across seasons, not this particular offering.

yea, could do for all time but i agree not ideal, i just think its a good stat to display more prominently, ill think on where best to put it. unless you think as is is ideal

Josh-Cena:
FWIW, I don't find this metric extremely useful, because lower percentage doesn't mean less credible scores; it's the _number_ that matters more. For the overview, we would need much more sophisticated statistics.

Josh-Cena:
@bearsyankees Good to merge?

bearsyankees:
Yeah LGTM, can always adjust later

bearsyankees (approved):

Issue #1633, opened 2024-03-11
More prominently remind users to do the challenge


Josh-Cena:
Now we don't require doing the challenge to access catalog data. This means users that aren't pre-verified (via Yalies) and haven't done the challenge are "logged in" but the catalog data still says "log in to see ratings". However, they don't know until they click on something else that needs evals access, such as the worksheet.

This is not good, since we are planning to make worksheets available to all users including those without evals access!

Instead, we should add an alert box telling users that "you are signed in but you need to do the challenge to access evals". I have three ideas:

1. It could be an item shown in the "me" dropdown, and we could add a little exclamation mark.
2. It could be a top-level banner.
3. It could be a popup in the bottom-right corner.

I slightly favor the first idea—because we don't want users to be able to dismiss it, otherwise it will be hard to bring it back up again.

benzuzu:
@Josh-Cena Thoughts on an actual button in this boxed whitespace area? I feel like usability wise it may be better to have something immediately visible rather than hidden in dropdown. I definitely see how options (2) and (3) are a bit too loud if we pop up on every refresh, but option (1) seems too subtle/hard to find.

<img width="1728" alt="Screenshot 2024-03-24 at 6 32 14 PM" src="https://github.com/coursetable/coursetable/assets/71607240/5ed0f658-9e79-4341-9f9b-fe5f64a97327">

Josh-Cena:
Hmmm. Maybe? I don't really know what to expect since like we are talking about a tiny fraction of our users.

PR #1639, opened 2024-03-14
feat: new auth ui

neilsong:


frontend/src/App.tsx
```diff
@@ -66,15 +66,26 @@ function App() {
...
-        courses with your family and friends with ease ;)
-      </Notice>
+      {location.pathname === '/catalog' && authStatus === 'unauthenticated' ? (
+        <Notice
+          // Increment for each new notice (though you don't need to change it
+          // when removing a notice), or users who previously dismissed the banner
+          // won't see the updated content.
+          key="unauthenticated-notice"
+          id={5}
+        >
```
Josh-Cena, line 76: Don't use a notice because it's closeable and then will never reopen. Use a separate dialog box in the lower right corner. See also https://github.com/coursetable/coursetable/issues/1633

Josh-Cena (changes requested):

frontend/src/App.tsx
neilsong: @benzuzu Thoughts on taking over? I'm kinda scuffed when it comes to frontend :)

neilsong: @Josh-Cena, I've thought about this more and I think we should replace the Me icon with a primary colored sign in button for unauthenticated access - something very similar to the approach google.com takes.

As for #1633, I think the exclamation point in the Me icon with an action item in the dropdown is the way to go.

Josh-Cena: Yep agreed for the first part. See my previous slack message:

> (For logged in users) I think it should show name initials instead of the default icon here, and we should have a way to let people customize their name in case there's no Yalies data, they have diff preferred name, etc.

But I also like your idea of a primary colored button.

bearsyankees:
@neilsong @Josh-Cena still viable?

Josh-Cena:
Yeah still viable. Not sure what exactly to do though.

Josh-Cena:
Closing as #1633 has been addressed and I think it's clear enough that users aren't logged in if they can't see ratings

PR #1641, opened 2024-03-15
refactor: ensure that x-hasura-role header is properly set

neilsong:
Problem: `http-proxy` and middleware request lifecycle handlers do not trigger with `bun`.
Solution: Avoid using the lifecycle callback


api/package.json
```diff
@@ -2,10 +2,9 @@
   "name": "api",
   "version": "0.0.0",
   "scripts": {
-    "generate": "bunx prisma generate",
     "start": "bun --hot --inspect=ws://0.0.0.0:6499/forcingPrefix src/server.ts",
     "start:prod": "bun src/server.ts",
-    "postinstall": "bun generate"
+    "postinstall": "bunx prisma generate"
```
neilsong, line 7: Changing this because the script isn't resolved properly when adding packages in `api`

api/src/server.ts
```diff
@@ -124,27 +124,29 @@ passportConfig(passport);
...
+// Add the authentication header to the request
+// Proxy initial HTTP requests to Ferry
+app.use('/ferry', (req, res, next) => {
+  req.headers['X-Hasura-Role'] = req.isAuthenticated()
+    ? 'student'
+    : 'anonymous';
+  const hasuraRole = req.headers['X-Hasura-Role'] ?? 'anonymous'; // Default to 'anonymous'
+  req.headers['X-Hasura-Role'] = hasuraRole;
+  ferryProxy(req, res, next);
+});
```
Josh-Cena, line 143:
```suggestion
app.use('/ferry', (req, res, next) => {
  const hasuraRole = req.isAuthenticated()
    ? 'student'
    : 'anonymous';
  req.headers['X-Hasura-Role'] = hasuraRole;
}, ferryProxy);
```

Something like this but with formatting

neilsong: Oh damn I didn't even know you could do this.

PR #1642, opened 2024-03-15
feat: API for generating link preview text

Josh-Cena:
Yet another go at https://github.com/coursetable/coursetable/pull/1630/

PR #1643, opened 2024-03-19
feat: og image 

bearsyankees:
branch of #1642, we can merge that one if its ready (or if vercel gets back to us) first, but should figure out how to approach the og image here. Bun does not support the canvas library yet but thats probs our best bet...


api/src/server.ts
```diff
@@ -1,35 +1,35 @@
+import * as Sentry from '@sentry/node';
```
Josh-Cena, line 1: Why are stuff moved around? Makes it hard to read the diff

bearsyankees: No idea, must have been from prettier or lint or some auto formatter. Nothing purposeful

Josh-Cena: Hmmm, yeah, looks like some sort of import sorting got run here.

bearsyankees:
eh we can re do this now

PR #1646, opened 2024-03-24
feat: make icons clickable in MeDropdown

benzuzu:
A bit of a pet peeve lol, but this does seem to be a webdev convention. We could also consider making the rows gray-hoverable.

See Reddit and Twitch for example.


Josh-Cena:
LGTM at a glance but afk right now, so will test&merge later

bearsyankees:
LGTM mergin thanks ben

PR #1650, opened 2024-03-29
feat: add friends on mobile (if we want)

bearsyankees:
prev could not add friends on mobile, but do we want this feature? and def could improve UI
Screenshot 2024-03-29 at 3 47 27 PM

### DIFF
--- a/frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
+++ b/frontend/src/components/Navbar/NavbarWorksheetSearch.tsx
@@ -181,7 +181,7 @@ function FriendsDropdown({
   );
 }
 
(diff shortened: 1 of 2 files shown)


Josh-Cena:
I think it's useful but UI needs a lot of work—the add friends dropdown is not parallel to the other three.

I think we shouldn't do this; instead, we should build a dedicated friends management page. This dropdown should be considered a convenience feature for desktop.

Josh-Cena:
I'm going to close this because I think a dedicated page is more useful on mobile. This dropdown is too hard to interact with on mobile and it's even harder to make it appear visually distinct from the other dropdowns.

PR #1653, opened 2024-04-01
refactor: fetch course info attributes from GQL

Josh-Cena:
Less need to update code


bearsyankees:
is this season dependent or all time? all time, right?

Josh-Cena:
All time

PR #1663, opened 2024-04-03
infra: prisma -> drizzle & mysql -> postgres

neilsong:
Huge infrastructure overhaul as title implies.

Drizzle now enables us to use TS to define the schema, which makes not only DB changes very simple but also helps local dev move fast.

Note: introspect does not lead to a clean TS schema for push, so we will use a hand-maintained schema file instead.

Before this PR is merged, the prod mysql db needs to be merged with ferry into a Postgres db. This will likely cause downtime, so a maintenance period must be scheduled

NEW DEV CHANGES:
 - Must run `npm run db:push` in `express` after change to `drizzle/schema.ts` to perform live DB migration
   - This includes the first time run!


bearsyankees:
do we want to do this before we migrate everything to postgres, in tandem, or does it not rly matter

Josh-Cena:
No they are orthogonal (I think, since we don't use weird DB primitives). I think this one is nicer to have and easier to do though.

api/postgres/.gitignore
```diff
@@ -0,0 +1 @@
+*data/
```
Josh-Cena, line 1: All `.gitignore` should be at the root

api/src/canny/canny.handlers.ts
```diff
@@ -59,13 +59,11 @@ export const cannyIdentify = async (
...
-      where: {
-        netId,
-      },
-      data: {
+    await db
+      .update(studentBluebookSettings)
+      .set({
         // Enable evaluations if user has a school code
-        evaluationsEnabled: Boolean(user.school_code),
+        evaluationsEnabled: user.school_code ? 1 : 0,
```
Josh-Cena, line 66: Ugh, does Drizzle not support booleans?

neilsong: I'll take a look into boolean definitions in the schema

neilsong:
## Status update

### Done
 - Prisma does not exist in our codebase
 - Basic worksheet manipulation works
 - All typechecking and linting passes

### Todo
 - Extended friends functionality needs to be tested - we'll need to do this in-person most likely
 - `docs/containers.md` and `docs/how-to-deploy.md` need to be updated

Josh-Cena:
> Extended friends functionality needs to be tested - we'll need to do this in-person most likely

You can manually insert friend requests/friends in the database, FWIW.

PR #1676, opened 2024-04-08
refactor: more efficient and refactored queries

neilsong:
Feel free to add more optimizations and refactors.


neilsong:
Currently old NetIDs without yalies data look like below. Should we filter or rename?

image

Josh-Cena:
> Should we filter or rename?

I think it's fine

api/src/friends/friends.queries.ts
```diff
@@ -0,0 +1,26 @@
...
+  const existingRequest = await tx
+    .selectDistinctOn([
+      studentFriendRequests.netId,
+      studentFriendRequests.friendNetId,
+    ])
+    .from(studentFriendRequests)
+    .where(
+      and(
+        eq(studentFriendRequests.netId, friendNetId),
+        eq(studentFriendRequests.friendNetId, netId),
```
Josh-Cena, line 21: This is quite confusing. We should name the params as `from` and `to` for this function

Josh-Cena:
@neilsong Please take another look; I've minimized the number of DB calls

api/src/friends/friends.handlers.ts
```diff
@@ -217,39 +164,29 @@ export const requestAddFriend = async (
...
-          eq(studentFriendRequests.friendNetId, friendNetId),
+        or(
+          and(
+            eq(studentFriendRequests.netId, netId),
+            eq(studentFriendRequests.friendNetId, friendNetId),
+          ),
+          and(
+            eq(studentFriendRequests.netId, friendNetId),
+            eq(studentFriendRequests.friendNetId, netId),
+          ),
```
neilsong, line 182: Is there ever a case where both exist? If this is the case, then the array destructuring won't work as intended

neilsong (commented):
LGTM

Josh-Cena (approved):

PR #1685, opened 2024-04-14
feat: Bootstrap v5

Josh-Cena:
This would enable us to use strict mode because react-bootstrap v1 doesn't support strict mode.


bearsyankees:
LGTM <3

Josh-Cena:
@bearsyankees one second

bearsyankees (approved):

Issue #1664, opened 2024-04-03
Race condition in API


Josh-Cena:
There are a lot of Sentry errors because someone could click the button twice in a row which causes DB write failure, or missing records, etc. Common examples:

- `NO_FRIEND_REQUEST`: caused by clicking accept friend req twice
- ``Unique constraint failed on the constraint: `netId_crn_season_worksheetNumber` ``: caused by adding the course twice (I think)

We should fix this on both backend and frontend:

- On backend, debounce all requests by (userID, payload)
- On frontend, change button into loading state until a response is received

PR #1686, opened 2024-04-15
feat: add load spinner to prevent clicking friends button twice

Josh-Cena:
As part of https://github.com/coursetable/coursetable/issues/1664, fixing the single most seen race condition


bearsyankees (commented):
LGTM

bearsyankees (approved):

PR #1709, opened 2024-07-22
feat: generate sitemap

bearsyankees:
basic form now perhaps will add on to create index etc later


package.json
```diff
@@ -1,31 +1,6 @@
 {
   "name": "coursetable-root",
-  "private": true,
```
Josh-Cena, line 3: ?

api/package.json
```diff
@@ -29,6 +29,7 @@
     "redis": "^4.6.14",
     "tsx": "^4.12.0",
     "winston": "^3.13.0",
+    "xmlbuilder2": "^3.1.1",
```
Josh-Cena, line 32: I strongly recommend using the `sitemap` package instead of hand-writing your own XML structure

package.json
bearsyankees: ya fax idk was auto

bearsyankees: oh it just moved it down

Josh-Cena: Yeah I'm not sure why the keys got reordered

api/src/catalog/sitemap.utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  await fs.writeFile(sitemapPath, xml, 'utf-8');
+
+  winston.info(`Sitemap generated for ${seasonCode} at ${sitemapPath}`);
+}
+
+export async function generateSitemaps(): Promise<void> {
+  let seasons: string[] = [];
+  try {
+    const sdk = getSdk(graphqlClient);
+    seasons = (await sdk.listSeasons()).seasons.map((x) => x.season_code);
```
Josh-Cena, line 48: No need to fetch twice—the `fetchCatalog` function already did both of these fetches. You can either do this function as part of `fetchCatalog`, or let `fetchCatalog` return the data you need, or even just read the JSON files that `fetchCatalog` wrote—your call.

api/src/catalog/sitemap.utils.ts
```diff
@@ -0,0 +1,66 @@
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import { Readable } from 'node:stream';
+import { SitemapStream, streamToPromise } from 'sitemap';
+import { getSdk, type CatalogBySeasonQuery } from './catalog.queries.js';
+import { STATIC_FILE_DIR, graphqlClient } from '../config.js';
+import winston from '../logging/winston.js';
+
+interface Course {
```
Josh-Cena, line 9: Call this `SitemapData` or `SitemapListing` etc. "Course" means something very specific in our terminology

package.json
Josh-Cena, line 3: You still need to revert the key reordering here

Josh-Cena (changes requested):

api/src/catalog/catalog.utils.ts
```diff
@@ -158,4 +220,18 @@ export async function fetchCatalog(overwrite: boolean) {
     ),
   );
   await Promise.all(processSeasons);
+
+  for (const seasonCode of seasons) {
```
Josh-Cena, line 224: Don't use await in a for loop if the loop isn't meant to be sequential. Parallelize it with `Promise.all`.

Also, try to reuse the data returned by `fetchData`—no need to fetch it again in this function. Some refactoring is probably necessary (for example you may want to actually do season sitemap generation as part of `fetchData`.

bearsyankees: thank u good call
should be better

PR #1712, opened 2024-07-24
docs: new blog post on link preview

Josh-Cena:
\*Excited noises\*


frontend/src/pages/releases/releases.tsx
```diff
@@ -10,6 +10,14 @@ type ReleaseNote = {
 };
 
 const releaseNotes: ReleaseNote[] = [
+  // Add more releases below
+  {
+    title: 'Optimizing Bot Traffic Handling for OG Tags: a Long Journey',
```
bearsyankees, line 15: How about "Optimizing Bot Traffic Handling for Link Previews: a Long Journey"

(may make it sound slightly more accessible for people who don't know what OG tags are) would change title in markdown file too obv

frontend/src/pages/releases/link-preview.mdx
```diff
@@ -0,0 +1,246 @@
...
+  alt="Slack conversation between Alex and Sida. Alex proposes to generate HTML the first time it is requested, but doubts if it would still be crazy."
+  width={600}
+/>
+
+Aside: You might ask, why don't we just change the URL to `https://coursetable.com/catalog/202403-10471`, so Vercel can serve the `catalog/202403-12309.html` file? Well there are still two problems: (1) there are probably hundreds of thousands of courses growing by several thousand every year; (2) course modals can be shown on other pages too, such as by clicking a course in your worksheet, which means… we have to have another set of HTML files that get served for `https://coursetable.com/worksheet/202403-10471`. That's a lot of HTML. We can't build so much HTML in reasonable time, and Vercel definitely has an upper bound to how much static assets we can store in it. This option was never considered or experimented with, but we did built infrastructure to statically pregenerate HTML for every URL at [#1536](https://github.com/coursetable/coursetable/pull/1536) (which Sida requested Alex's review on last year and he still hasn't…), in case it's useful for some future use case.
```
bearsyankees, line 95:
```suggestion
Aside: You might ask, why don't we just change the URL to `https://coursetable.com/catalog/202403-10471`, so Vercel can serve the `catalog/202403-12309.html` file? Well there are still two problems: (1) there are hundreds of thousands of courses growing by several thousand every year; (2) course modals can be shown on other pages too, such as by clicking a course in your worksheet, which means… we have to have another set of HTML files that get served for `https://coursetable.com/worksheet/202403-10471`. That's a lot of HTML. We can't build so much HTML in reasonable time, and Vercel definitely has an upper bound to how much static assets we can store in it. This option was never considered or experimented with, but we did built infrastructure to statically pregenerate HTML for every URL at [#1536](https://github.com/coursetable/coursetable/pull/1536) (which Sida requested Alex's review on last year and he still hasn't…), in case it's useful for some future use case.
```

we r def over 100k now i think (should update landing page too) but can do a quick query to check

frontend/src/pages/releases/link-preview.mdx
```diff
@@ -0,0 +1,246 @@
...
+- We also plan to further enhance our general handling of metadata to include information such as last modified ([#1619](https://github.com/coursetable/coursetable/issues/1619)), professor(s), related courses, and other relevant data, which is useful for search engine display.
+
+This experience has not only improved how CourseTable links appear on social media but also enables CourseTable courses to be indexed by search engines. For us, it provides valuable insights into our tech setup and experience for handling similar issues in the future. It is a great example of why we think CourseTable is so great. Through collaboration, creativity, and perseverance, we have made CourseTable more robust and user-friendly than ever before. We continue to fulfill our goal of making course selection and sharing as seamless and informative as possible for all Yale students—all while keeping it completely open and free.
```
bearsyankees, line 236: - indexed by search engines

a little redundant and not totally true yet wdyt

frontend/src/pages/releases/link-preview.mdx
Josh-Cena: Yeah looks like we are at 150,000

frontend/src/pages/releases/link-preview.mdx
Josh-Cena: It's true though! Without this, I don't think search engines can find those stuff because loading data takes time.

PR #1714, opened 2024-08-07
polish: add date and better search terms to titles

bearsyankees:
chose semi arbitrary dates for last updated though

### DIFF
--- a/api/src/catalog/catalog.utils.ts
+++ b/api/src/catalog/catalog.utils.ts
@@ -95,11 +95,18 @@ function transformListingToSitemapListing(
   listing: CatalogBySeasonQuery['listings'][number],
 ): SitemapListing {
   const year = parseInt(listing.season_code.substring(0, 4), 10);
+  const [today] = new Date().toISOString().split('T');
+  if (!today) throw new Error('Unable to determine the current date.');
+
   const season = listing.season_code.substring(4);
 
-  let lastmod = '';
-  if (season === '03' || season === '02') lastmod = `${year}-01-01`;
-  else lastmod = `${year - 1}-10-01`;
+  let semesterStartDate = '';
+  if (season === '01') semesterStartDate = `${year}-01-15`;
+  else if (season === '02') semesterStartDate = `${year}-05-25`;
+  else if (season === '03') semesterStartDate = `${year}-08-25`;
+  else throw new Error(`Unknown season code: ${listing.season_code}`);
+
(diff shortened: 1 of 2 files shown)


api/src/catalog/catalog.utils.ts
```diff
@@ -92,8 +94,16 @@ const exists = (p: string) =>
 function transformListingToSitemapListing(
   listing: CatalogBySeasonQuery['listings'][number],
 ): SitemapListing {
+  const year = parseInt(listing.season_code.substring(0, 4), 10);
+  const season = listing.season_code.substring(4);
+
+  let lastmod = '';
+  if (season === '03' || season === '02') lastmod = `${year}-01-01`;
```
Josh-Cena, line 101: Courses have never been published on 01-01; Maybe 04-01 is more realistic?

We can even choose roughly when the semesters start (01-15, 05-25, 08-25) because courses before this date usually change a lot.

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -182,8 +182,32 @@ function CourseModal() {
...
+        return { seasonName, seasonYear: year };
+      default:
+        throw new Error(`Unknown season code: ${seasonCode}`);
+    }
+  }
+
   if (!listing) return null;
-  const title = `${listing.course_code} ${listing.section.padStart(2, '0')} ${listing.course.title} | CourseTable`;
+  const { seasonName, seasonYear } = getSeasonDetails(listing.season_code);
+  const title = `${listing.course_code} ${listing.section.padStart(2, '0')}: ${listing.course.title} -- Yale ${seasonName} ${seasonYear} | CourseTable`;
```
Josh-Cena, line 210: We already have a `toSeasonString` util for this

api/src/catalog/catalog.utils.ts
bearsyankees: ya but im not sure we can list it as a course in the future like for the fall and spring 2024 classes we dont want those last mod in the future

Josh-Cena: What about "semester start date vs. today, whichever is earlier"?

bearsyankees: ya but then we want it constantly updating? i guess thats not horrible tbh....

Josh-Cena: Yeah doesn't sound like a bad idea since we only do this at most once a day anyway

PR #1715, opened 2024-08-19
feat: student reviews on landing page

bearsyankees:
Could be nice... (pulled from buymeacoffee)


bearsyankees:
@Josh-Cena thoughts

Josh-Cena:
As we discussed a while ago, we should at least provide a link to the original buymeacoffee page. Also maybe the two about "paying" aren't really appropriate.

bearsyankees:
@Josh-Cena look good 2 merge?

Josh-Cena (approved):
This is awesome let's go

Josh-Cena:
I think we should get a designer to remake the landing page. Looks a bit plain, WDYT?

PR #1728, opened 2024-09-27
fix: downgrade media query syntax

michaelcanudas:
## Fixed incompatible media query syntax

On some versions of Safari or older Chrome browsers, the following syntax is not supported:
```
@media (width >= 768px) { }
```

It instead just removes the CSS class, causing unexpected behavior on these browsers. This is a minor problem in most cases; however, for certain features (like the worksheet calendar view), it makes the UI unreadable.

This PR updates all instances (except in documentation) of media queries to use the syntax with better support:
```
@media (min-width: 768px) { }
```

This fixes the issue on most browsers; however, it does mean that we aren't using the most modern syntax when it is still supported. A future revision could be converting back to the more modern syntax and implementing an upstream check that would modify the CSS class to ensure support across all browsers.


Josh-Cena (commented):
1. You should still look at the ESLint errors even when they may not be directly introduced by your PR; try to fix them alongside. I think the reason you triggered these new ESLint errors is because our range syntax was unparsable by eslint-plugin-css-modules, so it didn't analyze those modules at all.
2. Unfortunately I don't think the `- 0.001px` trick works. On https://media-query-syntax.preview.coursetable.com/worksheet, try setting your screen to exactly `768px` wide. You'll find the sidebar appearing at the bottom but with desktop width.

   image
   
   I think the reason is that screen widths have a max precision. I recommend trying with `- 0.3px` instead, but we may have to go as coarse as `- 0.5px`. You can play with it locally.

Josh-Cena (approved):
Great, thanks!

PR #1732, opened 2024-10-01
chore: add user survey

bearsyankees:
@Josh-Cena do we want to add any click trackers or anything


bearsyankees:
@Josh-Cena imma be offline but feel free to merge whenever/add sentry crumb

PR #1733, opened 2024-10-05
feat: add section dropdown to course modal

### LINKED ISSUE #1694: Link to cross-listings and other sections in the course modal
In the course modal header, the course codes should be clickable and take me to that cross listing in the same season. The section number should also be in the header with a dropdown to let me select other sections in the same season.

https://coursetable.canny.io/admin/board/feature-requests/p/link-to-cross-listings-and-other-sections-in-the-course-modal

wesleyandrade05:
## Pre-flight checklist

- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Summary

This PR implements the following **features**

- Section dropdown on modal header

## Test plan (required)

https://github.com/user-attachments/assets/05e7988a-ea8f-4702-bde2-c4984890b9c1

## Related issues/PRs

Fixes #1694


wesleyandrade05:
Part of #1694

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -181,6 +183,20 @@ function CourseModal() {
       setHistory([listingFromQuery]);
     });
   }, [history.length, searchParams, requestSeasons, courses]);
+  useEffect(() => {
+    const crossSections: CourseModalHeaderData[] = [];
+    const courseModal = searchParams.get('course-modal');
+    if (!courseModal) return;
+    const seasonCode = courseModal.split('-')[0] as Season;
+    courses[seasonCode]?.forEach((course: CourseModalHeaderData) => {
```
Josh-Cena, line 191: @wesleyandrade05 So this may be a bit confusing, but `courses[seasonCode]` does not always exist. It only exists if the course modal is opened from the catalog, or otherwise the corresponding JSON happens to be available because the user has viewed the catalog/worksheet of that semester. If you navigate to this modal using the past course links, then the full catalog will not exist because this modal data is acquired from GraphQL not from the static JSON catalog. Therefore I think you should always request fresh data from GraphQL instead.

wesleyandrade05: You mean using the `requestSeasons` function?

Josh-Cena: No. That one downloads the full JSON data. We just want to be able to get the related course data about all other sections. In fact all we need is the CRN.

wesleyandrade05: Got it. Where can I see what is the specific API I would need to query? I am a bit lost on that. I see this is what gets the full catalog:
```
const res = await fetchAPI(`/catalog/public/${season}`, {
    breadcrumb,
  });
```

But I can't see where I can just fetch graphQL for a specific coursename, for example (to find the sections)

Josh-Cena: See how we fetch the course modal info for example. It's quite complicated actually and we probably benefit from some docs, but TLDR there's no separate endpoint for it. You write a GQL query in the GQL query under queries, then run codegen, and then import the generated SDK. Take a look at how the other queries there are used.

bearsyankees:
Ur the goat, thanks for taking initiative on this.

for some classes (like history of south africa) just seeing section 01 (even tho sections exist), . and i think we also maybe we show a popup when adding a lecture to make sure they select a section? also, what do we think on showing the section times in the dropdown or somehow showing what times the sections are?

this is epic and on the right path i bet shippable in a week

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -181,6 +183,22 @@ function CourseModal() {
...
+    void requestSeasons([seasonCode]).then(() => {
+      courses[seasonCode]?.forEach((course: CourseModalHeaderData) => {
+        if (course.course_code === listing?.course_code)
+          crossSections.push(course);
+      });
+      setSections(
+        crossSections.sort((a, b) => a.section.localeCompare(b.section)),
+      );
+    });
+  }, [listing, courses, searchParams, requestSeasons]);
```
bearsyankees, line 200: also I think here we can make a new gql query instead of requesting the whole catalog right @Josh-Cena

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -181,6 +183,22 @@ function CourseModal() {
...
   }, [history.length, searchParams, requestSeasons, courses]);
+  useEffect(() => {
+    const crossSections: CourseModalHeaderData[] = [];
+    const courseModal = searchParams.get('course-modal');
+    if (!courseModal) return;
+    const seasonCode = courseModal.split('-')[0] as Season;
+    void requestSeasons([seasonCode]).then(() => {
+      courses[seasonCode]?.forEach((course: CourseModalHeaderData) => {
+        if (course.course_code === listing?.course_code)
+          crossSections.push(course);
```
bearsyankees, line 194: also I think here we can make a new gql query instead of requesting the whole catalog right @Josh-Cena

Josh-Cena: Yep that's what I suggested above

wesleyandrade05:
@bearsyankees what is weird about this history of south africa is that I am not even seeing the sections when I search for them. Maybe this second catalog query that I was doing is bugging something (for example, in my local host, I see the past courses fine on the modal, but on the deployed website here they are all messed up). I will see how that goes once I make the change to just make a new gql query.

I like the idea about the popup and the meeting times on the dropdown. However, don't you think the popup should have its own PR?

Josh-Cena:
Here are a few more feedbacks.

1. We should show at least the section titles in the select because not every section has the same title. Better yet we should show the time and professor too. All this can be easily done with a GQL query.
2. Instead of using the browser's native `<select>`, use our custom select component, which allows more complex typesetting.

bearsyankees:
@wesleyandrade05 this looks preeeettty fire to me, i guess just would we want a second PR to confirm they realize they are adding a section? and do we still want to have show discussion sections as an advanced filter? i kinda think yea if we r just gonna display them the way this PR does so people can look for conflicts. lets discuss at meeting

bearsyankees:
make dropdown more visible + dropshadow + label + no text input + remove colon if no times

bearsyankees (changes requested):
(see above comment)

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -198,11 +198,6 @@ function CourseModal() {
     });
   }, [history.length, searchParams, requestSeasons, courses]);
 
-  useEffect(() => {
-    if (!data) return;
-    setSections(data.listings as Listings[]);
-  }, [data, courseCode, season]);
-
```
Josh-Cena: https://react.dev/learn/you-might-not-need-an-effect

frontend/src/components/CourseModal/CourseModal.tsx
```diff
@@ -255,7 +250,7 @@ function CourseModal() {
       >
         <PopoutSelect<Option, false>
           value={sectionsOptions.get(listing.section)}
-          options={sectionsOptions.values().toArray()}
+          options={[...sectionsOptions.values()]}
```
Josh-Cena, line 253: Check https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/toArray; its support is too low, so you need to either install a polyfill from core-js (I don't think Vite automatically polyfills) or just use `[...]` which I did here

wesleyandrade05:
@Josh-Cena can we remove those extra fields? I left them there because I need to have the whole course information to add it into the listing history. I was having some bugs when not putting all the fields.

Josh-Cena:
There's a layout shift when you open the course modal. This is because the section dropdown is taller than the course codes, so when the sections get fetched and displayed, that row gets expanded slightly. I think we should render a dropdown of size 1 by default, and "hydrate" it with extra data once we have them, instead of rendering `null` by default.

wesleyandrade05:
> @Josh-Cena can we remove those extra fields? I left them there because I need to have the whole course information to add it into the listing history. I was having some bugs when not putting all the fields.

@Josh-Cena  just bringing that up again. Have you considered this? (not sure if maybe you changed something else that would fix it)

wesleyandrade05:
> There's a layout shift when you open the course modal. This is because the section dropdown is taller than the course codes, so when the sections get fetched and displayed, that row gets expanded slightly. I think we should render a dropdown of size 1 by default, and "hydrate" it with extra data once we have them, instead of rendering `null` by default.

I am not sure if I understood this. What do you mean by layout shift?

Josh-Cena:
Actually I already fixed this in https://github.com/coursetable/coursetable/pull/1733/commits/2adaba02c41a7e26939b3661b3385d714a18e83d. Don't worry; I'll take care of any other feedback I may have.

bearsyankees:
This seems good to merge

Josh-Cena (approved):

Josh-Cena:
:tada:

PR #1734, opened 2024-10-09
fix: filter graduate course using school

maydeline:
Changing the filtering logic function of isGraduate from filter by the course number to the school.

### DIFF
--- a/frontend/src/utilities/course.ts
+++ b/frontend/src/utilities/course.ts
@@ -360,8 +360,8 @@ type CourseWithEnrolled = {
   last_enrollment_same_professors?: boolean | null;
 };
 
-export function isGraduate(listing: Pick<Listings, 'number'>): boolean {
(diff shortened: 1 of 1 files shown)


frontend/src/utilities/course.ts
```diff
@@ -360,8 +360,8 @@ type CourseWithEnrolled = {
   last_enrollment_same_professors?: boolean | null;
 };
 
-export function isGraduate(listing: Pick<Listings, 'number'>): boolean {
-  return Number(listing.number.replace(/\D/gu, '')) >= 500;
+export function isGraduate(listing: Pick<Listings, 'school'>): boolean {
+  return listing.school !== "YC";
```
Josh-Cena, line 364:
```suggestion
  return listing.school !== 'YC';
```

Josh-Cena (approved):
Great, thanks!

PR #1742, opened 2024-11-07
refactor: rename endpoint to "updateWorksheetCourses"

Etherite1:
Changed occurrences of "updateWorksheet" to "updateWorksheetCourses" for better naming convention


Josh-Cena:
Is this change made because we will have an API that makes meta-changes to the worksheets as a whole?

Etherite1:
> Is this change made because we will have an API that makes meta-changes to the worksheets as a whole?

Yeah, I thought it'd be apt to call the updates to worksheet metadata "updateWorksheet" rather than updates to courses in the worksheet

Josh-Cena (approved):

PR #1747, opened 2024-11-11
feat: new worksheetNames that tracks all non-main worksheets

Etherite1:
- New schema that uniquely maps (netId, season, worksheetNumber) to worksheetName
- When adding/removing/updating courses, worksheetNames table updates also

Need: DB migration that adds all the current worksheets

TODO: New handlers for add/rename/delete worksheets


neilsong:
A few minor things you might want to think about:

- I know we define our text columns as varchar 64 but perhaps we want to start migrating this to `text()` instead. Overall just a better sql type.

- Should the worksheet name be unique? (If so, drop a `.unique()` constraint on the column)

- Do we want to auto-delete worksheets when they empty? Unintended side effects:
	1. I added x course to new worksheet
	2. I remove it, wanting to add some other course instead afterwards
	3. The worksheet is gone

- I think we can resolve worksheet management explicitly, like how friends currently works

Etherite1:
> A few minor things you might want to think about:
> 
> * I know we define our text columns as varchar 64 but perhaps we want to start migrating this to `text()` instead. Overall just a better sql type.
> * Should the worksheet name be unique? (If so, drop a `.unique()` constraint on the column)
> * Do we want to auto-delete worksheets when they empty? Unintended side effects:
>   
>   1. I added x course to new worksheet
>   2. I remove it, wanting to add some other course instead afterwards
>   3. The worksheet is gone
> * I think we can resolve worksheet management explicitly, like how friends currently works

Not sure whether worksheet name should be unique - open to ideas but I think users should be able to name worksheets the same if they want. As for the auto-delete - what Sida and I agreed on is to have this for now while we don't push frontend changes (to add/remove worksheets). This will make it so that when we do have the extra functionality, we don't store worksheets that the user didn't create and that are empty.

Can you clarify what you mean by resolving worksheet management thx

Josh-Cena:
> Should the worksheet name be unique? 

No because we can't enforce that upon our users, especially since all new worksheets will be called "New worksheet".

> Do we want to auto-delete worksheets when they empty? Unintended side effects:

This is only as a temporary step before we have the full CRUD UI.

api/src/user/user.handlers.ts
```diff
@@ -78,6 +79,29 @@ async function updateWorksheetCourse(
...
+    ])
+    .from(worksheetNames)
+    .where(
+      and(
+        eq(worksheetNames.netId, netId),
+        eq(worksheetNames.season, season),
+        eq(worksheetNames.worksheetNumber, worksheetNumber),
+      ),
+    );
+  if (!nameExists && worksheetNumber > 0) {
```
Josh-Cena, line 96: I suggest this:

```js
if (worksheetNumber > 0) {
  const [nameExists] = ...
  if (!nameExists) { ... }
}
```

api/src/user/user.handlers.ts
```diff
@@ -78,6 +79,31 @@ async function updateWorksheetCourse(
...
+        ),
+      );
+    if (!nameExists) {
+      await db.insert(worksheetNames).values({
+        netId,
+        season,
+        worksheetNumber,
+        worksheetName: `Worksheet ${worksheetNumber}`,
+      });
+    }
```
Josh-Cena, line 104: I _think_ you can refactor this into a single insert with a conflict-do-nothing, but I'll let @neilsong take look

neilsong (approved):

PR #1748, opened 2024-11-12
feat: backend handlers for new worksheetName table

Etherite1:
- updateWorksheetNames handler: adds worksheet (finds MEX of existing worksheet numbers), delete/rename worksheet
- getWorksheetNames handler: mapping from [season][worksheetNumber] -> worksheetName
- when switching between friends or back to yourself, goes back to main worksheet
- frontend API fetches


api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,167 @@ export const getUserWorksheet = async (
...
+          eq(worksheetMetadata.season, season),
+        ),
+      )
+      .orderBy(worksheetMetadata.worksheetNumber);
+
+    const worksheetNumbers = worksheetNumbersRes.map(
+      (row) => row.worksheetNumber,
+    );
+
+    const firstAvailableWsNumber = getFirstAvailableWsNumber(worksheetNumbers);
```
Josh-Cena, line 312: Don't you need to return this number to client side? Or are you just gonna reload and refetch everything

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,167 @@ export const getUserWorksheet = async (
     data: worksheetCoursesToWorksheets(worksheets)[netId] ?? {},
   });
 };
+
+const AddWorksheetSchema = z.object({
+  action: z.literal('add'),
+  season: z.string().transform((val) => parseInt(val, 10)),
```
Josh-Cena, line 256: Can we take a "name" parameter? I'm thinking the following UI: the bottom of the worksheet names list is a "+", and when you click it, it turns into a text box, which by default is filled with "New worksheet", and you can edit this text before you even request adding it. This means the API doesn't have to hardcode this default name.

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,167 @@ export const getUserWorksheet = async (
...
+    const [existingWorksheet] = await db
+      .select()
+      .from(worksheetMetadata)
+      .where(
+        and(
+          eq(worksheetMetadata.netId, netId),
+          eq(worksheetMetadata.season, season),
+          eq(worksheetMetadata.worksheetNumber, worksheetNumber),
+        ),
+      );
```
Josh-Cena, line 331: I'm pretty sure you can do it within one query to avoid race conditions. If you add `.returning();` at the end it will return the deleted object(s), and you can return an error if the returned list is empty.

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,167 @@ export const getUserWorksheet = async (
...
+  } else {
+    const { worksheetNumber, worksheetName } = bodyParseRes.data;
+    const [existingWorksheet] = await db
```
Josh-Cena, line 350: Same here

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,167 @@ export const getUserWorksheet = async (
...
+    })
+    .from(worksheetMetadata)
+    .where(eq(worksheetMetadata.netId, netId));
+
+  winston.info(
+    `Retrieved ${allWorksheetMetadata.length} worksheets for user ${netId}`,
+  );
+
+  const worksheetMap: {
+    [season: string]: { [worksheetNumber: number]: string } | undefined;
```
Josh-Cena, line 405: This doesn't follow your documented API shape—the API you documented is `{ [worksheetNumber: number]: { worksheetName: string } }` which IMO is much better

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,167 @@ export const getUserWorksheet = async (
...
+  } = {};
+
+  allWorksheetMetadata.forEach(({ season, worksheetNumber, worksheetName }) => {
+    worksheetMap[season] ??= {};
+    worksheetMap[season][worksheetNumber] = worksheetName;
+  });
+  res.json({
+    netId,
+    metadataForWorksheet: worksheetMap,
```
Josh-Cena, line 414:
```suggestion
    worksheets: worksheetMap,
```

I guess... Less words.

api/src/user/user.utils.ts
```diff
@@ -33,3 +33,16 @@ export function worksheetCoursesToWorksheets(
   }
   return res;
 }
+
+export function getFirstAvailableWsNumber(worksheetNumbers: number[]) {
```
Josh-Cena, line 37: If we are returning the worksheets in the order of this number, then they should be numbered in insertion order, so I suggest just an incrementing counter without MEX.

docs/api.md
```diff
@@ -382,6 +382,61 @@ Endpoints marked as "needs eval access" additionally returns 401 with `error: "U
     };
     ```
 
+## WorksheetMetadata
```
Josh-Cena, line 385: I think this should be merged with the previous section since it's still about "worksheets" (in general we put all `/api/x/...` under the same H2)

api/src/user/user.utils.ts
Etherite1: This makes sense!

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,157 @@ export const getUserWorksheet = async (
...
+    .where(eq(worksheetMetadata.netId, netId));
+
+  winston.info(
+    `Retrieved ${allWorksheetMetadata.length} worksheets for user ${netId}`,
+  );
+
+  const worksheetMap: {
+    [season: string]:
+      | { [worksheetNumber: number]: { worksheetName: string } }
+      | undefined;
```
Josh-Cena, line 395: When can it be undefined? I don't think you need `| undefined`

api/src/user/user.utils.ts
```diff
@@ -33,3 +33,10 @@ export function worksheetCoursesToWorksheets(
   }
   return res;
 }
+
+export function getNextAvailableWsNumber(worksheetNumbers: number[]): number {
+  // Returns max(...worksheetNumbers) + 1, 1 if empty
```
Josh-Cena, line 38: You can remove this comment

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,157 @@ export const getUserWorksheet = async (
...
+    const nextAvailableWsNumber = getNextAvailableWsNumber(worksheetNumbers);
+
+    await db.insert(worksheetMetadata).values({
+      netId,
+      season,
+      worksheetNumber: nextAvailableWsNumber,
+      worksheetName,
+    });
+    res.json({
+      newWsNumber: nextAvailableWsNumber,
```
Josh-Cena, line 323: Suggestion: call it `worksheetNumber` to make the naming consistent

docs/api.md
```diff
@@ -382,6 +382,60 @@ Endpoints marked as "needs eval access" additionally returns 401 with `error: "U
...
+    - `season`: `string`
+    - `worksheetNumber`: `number`
+    - `worksheetName`: `string`
+
+#### Response
+
+**Status: 200**
+
+- If `action` == `"add"`:
+  - returns the worksheet number assoc. with the new worksheet
```
Josh-Cena, line 409: Need to say it's `{ worksheetNumber: number }`

docs/api.md
```diff
@@ -382,6 +382,60 @@ Endpoints marked as "needs eval access" additionally returns 401 with `error: "U
...
 
+### `POST` `/api/user/updateWorksheetMetadata`
+
+#### Request
+
+- Needs credentials
+- Body:
+  - Option 1 (add):
+    - `action`: `"add"`
+    - `season`: `string`
```
Josh-Cena, line 393: Need to update this

frontend/src/queries/api.ts
```diff
@@ -463,6 +499,22 @@ export async function fetchUserWorksheets() {
...
+export async function fetchUserWorksheetMetadata() {
+  const res = await fetchAPI('/user/worksheetMetadata', {
+    breadcrumb: {
+      category: 'worksheet',
+      message: 'Fetching user worksheet names',
+    },
+    schema: z.object({
+      netId: netIdSchema,
+      worksheets: z.record(z.string(), z.record(z.string(), z.string())),
+      // { [season]: { [worksheetNumber]: worksheetName } }
```
Josh-Cena, line 511: Need update

Josh-Cena (approved):
A few nits, then let's try! I'll update the prod table's name.

api/src/user/user.handlers.ts
```diff
@@ -247,3 +250,156 @@ export const getUserWorksheet = async (
...
+  winston.info(
+    `Retrieved ${allWorksheetMetadata.length} worksheets for user ${netId}`,
+  );
+
+  const worksheetMap: {
+    [season: string]: { [worksheetNumber: number]: { worksheetName: string } };
+  } = {};
+
+  allWorksheetMetadata.forEach(({ season, worksheetNumber, worksheetName }) => {
+    worksheetMap[season] ??= {};
```
Josh-Cena, line 397:
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
    worksheetMap[season] ??= {};
```

PR #1750, opened 2024-11-12
fix: duplicated schema unique indexing

Etherite1:
whoops!


Josh-Cena:
OK Neil already fixed in https://github.com/coursetable/coursetable/commit/40562f4daa82b2e0e4f079f71508b1ea39314555 but thanks

PR #1751, opened 2024-11-13
feat: add PWA installation

abheekda1:
Implements PWA features


abheekda1:
I left it as a an actual PR rather than a draft because I don't know if a draft will get its own staging build.

PR #1758, opened 2024-11-15
feat: PWA installation

abheekda1:
- [x] Ability to install as a mobile app (needs more testing, esp w Android)
- [x] Ability to install as a desktop app
- [ ] Offline caching
- [ ] Other features (notifications, etc.)


bearsyankees:
this is huge

frontend/index.html
bearsyankees, line 1: manifest supposed to be commented out?

frontend/src/main.tsx
```diff
@@ -10,11 +10,11 @@ import {
 import * as Sentry from '@sentry/react';
 
 // Globals has to be imported first, because it contains all the base CSS!
-// eslint-disable-next-line import/order
-import Globals from './Globals';
+
```
bearsyankees, line 13: is this previous comment irrelevant or should we keep globals first, and can j keep the eslint-disable? but maybe works regardless now

frontend/vite.config.ts
```diff
@@ -115,6 +116,27 @@ export default defineConfig({
...
+      },
+      injectRegister: 'auto',
+      manifest: {
+        name: 'CourseTable',
+        short_name: 'CourseTable',
+        start_url: '/',
+        icons: [
+          {
+            src: 'icon200x200.png',
+            // Sizes: '200x200',
```
bearsyankees, line 132: comments here needed?

package.json
bearsyankees, line 1: why the order switch ? maybe accidental or needed to pass ci?

frontend/src/main.tsx
Josh-Cena: It doesn't work; the preview's styles are literally broken 😅

bearsyankees: ya i hadn't looked since the changes to pass ci lmao, def need to revert this

frontend/index.html
abheekda1: It's not supposed to be there at all actually

bearsyankees: ya feel free to j remove then

frontend/src/main.tsx
abheekda1: Idk how this happened I think it had something to do with me running eslint --fix at some point

frontend/vite.config.ts
abheekda1: nope

package.json
abheekda1: Yeah accidental, some automatic thing?

abheekda1:
I think automatic dark mode can also be implemented later with
```css
@media (prefers-color-scheme: dark) {
  :root:not([data-theme]) {
    ...
  }
  ```

File: frontend/src/main.tsx [hunk 3]
```diff
@@ -12,6 +12,7 @@ import * as Sentry from '@sentry/react';
 // Globals has to be imported first, because it contains all the base CSS!
 // eslint-disable-next-line import/order
 import Globals from './Globals';
+
```
Josh-Cena, line 15: Remove this line

package.json
```diff
@@ -1,32 +1,6 @@
 {
   "name": "coursetable-root",
-  "private": true,
```
Josh-Cena, line 3: Need to revert the ordering change

frontend/vite.config.ts
```diff
@@ -63,7 +64,7 @@ function remarkPluginAddHeadingId(): Transformer {
 
       if (!parsedId) return;
       // When there's an id, it is always in the last child node
-      // Sometimes heading is in multiple "parts" (** syntax creates a child
+      // Sometimes heading is in multiple 'parts' (** syntax creates a child
```
Josh-Cena, line 67: Revert

Josh-Cena (approved):
Copying over my feedback:

The code looks good to me and I think we can try it out like this. But if it matters: on iPad, the user agent string does not contain "ipad"; there's absolutely nothing telling iPad from mac Safari. I'm reading https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable and it doesn't seem like there's a good way to detect whether browser supports PWA—so I'm fine with only showing this button on iPhone

PR #1761, opened 2024-11-15
feat: user API changes

Etherite1:
- lots of new typing and refactoring
- moved worksheet metadata into getWorksheets handler, removed getWorksheetMetadata handler
- worksheet name and courses in that worksheet are now grouped together


Josh-Cena (approved):
Great, thank you!

Issue #1703, opened 2024-07-14
Profile page


Josh-Cena:
We should have a profile page that allows users to customize and/or hide data.

Currently users have the following data in the DB (not everything's used for display):

- First/last name
- Net ID
- Email
- UPI
- School
- Evaluations enabled
- Year
- College
- Major/Curriculum

We should allow:

- When searching for friends, allow each name to be clickable so you can see certain info of others (a la Yalies)
- Be able to customize your own info, such as changing your preferred first/last name, or fill them in if they don't exist in DB
- Set certain fields, such as major, to be hidden
- Enter the challenge page from this page, if eval access is false
- Be able to revoke evals access? (Useful for testing the challenge, for us)
- Browse through your friends list, in addition to the worksheet page, and edit the list (add, remove, etc.)

Then, we can consider changing the profile icon to be your own initials rather than a generic person icon.

neilsong:
Can also consider syncing user settings such as light/dark mode, default filters, etc. in enhancement PR with profile icon customization

neilsong:
Make sure to include recently added global preferences such as "Group by professors", "Separate curves for each course", and "Group recurring courses"

Josh-Cena:
FWIW, I see these toggles as "quick switches" rather than some global state. I don't think there's a global preference of one over the other, and syncing it to the server would be too costly at little benefit.

PR #1767, opened 2024-11-20
feat: display user name initials in profile icon

Josh-Cena:
1. The name is already visible for friends, so we should make the user know their own name as stored in the DB.
2. More easily know if you are logged in.
3. Part of https://github.com/coursetable/coursetable/issues/1703


neilsong:
Will `Boolean("")` return true?

Josh-Cena:
`Boolean()` uses the same falsy/truthiness check as `if ()` etc. So `""` is also `false`. I don't believe this can happen anyway.

bearsyankees:
LGTM, merge?

PR #1768, opened 2024-11-22
feat: random course button

bearsyankees:
raandddoomm


bearsyankees:
v1.1, not responsive to filters

Josh-Cena:
@bearsyankees Rerun with `./start.sh -f -d` to fetch the latest Ferry schema

Josh-Cena:
Does not look good at a very particular screen width:

image

We should make the button have the same look and feel as the reset button (but different color).

Also I suggest moving it here:

Screenshot 2024-11-23 at 02 01 21

frontend/src/components/Search/RandomButton.module.css
```diff
@@ -0,0 +1,15 @@
+.randomButton {
+  font-size: 70%;
+  font-weight: 600;
```
Josh-Cena, line 3:
```suggestion
  font-weight: 500;
```

Josh-Cena (approved):
:fire:

PR #1773, opened 2024-12-02
feat: add, remove, rename worksheets

Etherite1:
Add, remove, rename worksheets. New logic to handle friends and currently viewed worksheets.
- Temporary API changes removed
- New UI for selecting worksheets/renaming/adding/deleting
- Main worksheet exists in DB iff it has nonzero courses
- Refactors, differentiating between "my viewed worksheet number" and "viewed worksheet number". Former is used for the one-click add, and latter is used for viewing purposes

TODO: Name of worksheet overflows onto buttons if too long

Fix https://github.com/coursetable/coursetable/issues/1457

Testing:
- Add worksheet
- Add worksheet with same name as curr worksheeet
- Add worksheet with "Main Worksheet"

- Rename not currently selected worksheet
- Rename currently selected worksheet

- Delete currently selected worksheet
- Delete not currently selected worksheet

- Select main worksheet, switch to friend, switch to one of their non-main worksheets, check still adding to main worksheet
- Select non main worksheet, switch to friend, switch to one of their non-main worksheets, check still adding to earlier selected non-main worksheet

- Course modal WS dropdown - selecting one of those still adds to correct worksheet
- Course modal WS dropdown - selecting one of those doesn't change the myCurrentlyViewedWS

- Change season - reverts to main worksheet, everything adds correctly, etc
- Adding course from friend's worksheet - adds to myCurrentlyViewedWorksheet
- Clicking out of friend's worksheets back to "me" - toggles to myCurrentlyViewedWorksheet
- Friend deletes worksheet that you're currently viewing, and you refresh - go to their main worksheet
- Remove course while viewing friend's worksheet that you both have - removes correctly from your worksheet, not theirs


frontend/src/slices/UserSlice.ts
```diff
@@ -88,4 +92,34 @@ export const createUserSlice: StateCreator<Store, [], [], UserSlice> = (
       toast.info(`Sent friend request: ${friendNetId}`);
     }
   },
+  async addWorksheet(season: Season, name: string) {
```
Josh-Cena, line 95: FYI you don't need to add these API handles to the user context unless you need something about the user. The friends features are there because we do some client validations before we issue the request, but you don't have those (or at least not yet), so you can directly call the API functions.

Etherite1: Amazing, thanks

frontend/src/components/Worksheet/WorksheetNumberDropdown.tsx
```diff
@@ -1,19 +1,200 @@
...
+  enterAction,
+  onCancel,
+}: {
+  readonly startingInput: string;
+  readonly enterAction: (newWsName: string) => void;
+  readonly onCancel: () => void;
+}) {
+  const [inputField, setInputField] = useState(startingInput);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [cursorPos, setCursorPos] = useState(0);
```
Josh-Cena, line 28: What's this doing exactly? Does the native cursor management not work?

Etherite1: Yeah there's some issues with selection in the textbox (like selecting a chunk of text, or the cursor position persisting after the input textbox is modified) - this was the best I could come up with

Josh-Cena: @Etherite1 I tried removing it and the behavior appears the same. Could you try again and make a recording if it doesn't work?

Etherite1: Seems to be okay now. Not sure why it didn't work before... thanks for the catch!

Josh-Cena (approved):
And here it goes!

PR #1774, opened 2024-12-04
fix: api bugs

Etherite1:
- .returning() can't be empty
- due to foreign key relationship, must delete all courses with that worksheet id before deleting the worksheet's metadata


Josh-Cena (approved):
:+1:

PR #1780, opened 2024-12-23
feat: allow for offline dev

bearsyankees:
Basically just use docker cache, now that we use DB snapshots we can dev offline

didn't touch doppler bc it uses fallback automatically if offline but maybe a cleaner way to do that to avoid the error message? but i think it's fine


neilsong:
What you can do simplify this a little is set the `--pull-always` flag as a variable, so that it's empty when `$OFFLINE == false`, similar to the `$FORCE_RECREATE`

neilsong (commented):
What you can do simplify this a little is set the `--pull-always` flag as a variable, so that it's empty when `$OFFLINE == false`, similar to the `$FORCE_RECREATE`

bearsyankees:
> What you can do simplify this a little is set the `--pull-always` flag as a variable, so that it's empty when `$OFFLINE == false`, similar to the `$FORCE_RECREATE`

this is why we pay you the big bucks

PR #1790, opened 2025-01-13
feat: private worksheets

Etherite1:
Adding worksheet settings for worksheets, can make worksheet private (hidden to friends)


api/drizzle/schema.ts
```diff
@@ -94,6 +94,7 @@ export const worksheets = pgTable(
     season: integer('season').notNull(),
     worksheetNumber: integer('worksheetNumber').notNull(),
     name: varchar('name', { length: 64 }).notNull(),
+    isPrivate: boolean('isPrivate').notNull().default(false),
```
Josh-Cena, line 97: I prefer just `private`

Etherite1: I got some linter error on the frontend saying something about private being a protected keyword, so I switched

Josh-Cena: Yeah you can't use it as an identifier, but you can use it as an object property

Etherite1: Sounds good, I switched it to private

Etherite1: Let me know if this looks good to merge

frontend/src/components/Worksheet/WorksheetCalendarList.tsx
```diff
@@ -120,6 +155,43 @@ function WorksheetCalendarList() {
...
+              onChange={async () => {
+                await updateWorksheetMetadata({
+                  season: viewedSeason,
+                  action: 'setPrivate',
+                  worksheetNumber: viewedWorksheetNumber,
+                  private: !isViewedWorksheetPrivate,
+                });
+                await worksheetsRefresh();
+              }}
+              disabled={viewedWorksheetNumber === 0}
```
Josh-Cena, line 182: Add a tooltip when it's disabled explaining why

Etherite1: done!

PR #1797, opened 2025-01-26
feat: location filter

henzheng:
Adds the feature under the advanced panel to filter for certain buildings or locations.


Josh-Cena:
This should be like info attributes, so instead of creating a new folder in `api`, you just need to create a new query under catalog queries.

frontend/src/contexts/searchContext.tsx
```diff
@@ -93,6 +95,15 @@ export const schoolsOptions = Object.entries(schools).map(
   }),
 );
 
+const buildings = buildingData as Buildings[];
+
+export const buildingOptions = buildings.map(
+  (building): Option => ({
+    value: building.code,
+    label: building.building_name ?? building.code,
```
Josh-Cena, line 103: I suggest doing `${building.building_name} (${building.code})` if the name exists, because the code is still useful in many contexts

Josh-Cena:
```suggestion
    label: building.building_name
      ? `${building.building_name} (${building.code})`
      : building.code,
```

(Formatting needed

Josh-Cena (approved):
Nice work, thanks!

PR #1807, opened 2025-02-23
fix: can't change hidden state from friend worksheet

Etherite1:
not sure why I didn't catch this bug earlier 🤔


Josh-Cena:
Branch messed up?

Etherite1:
> Branch messed up?

Yeah not sure why. It's the same thing where prettier is removing stuff from graphql

Josh-Cena:
Probably want to pull latest master and rebase

Etherite1:
Yeah I ended up branching off origin/master which was fine

PR #1816, opened 2025-03-25
Add support for 4 digit codes

michaelcanudas:


bearsyankees (approved):

bearsyankees:
could have a single label-generating function that takes both the value and a second parameter for top or bottom but idrc @neilsong probs fine w 2 label props

PR #1838, opened 2025-10-25
ci fixes

reybahl:
- fix build issues by adding `@popperjs/core` as an explicit dependency
- fix typescript/MDX configuration with better type annotations


frontend/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "build": "vite build",
+    "build": "bun run vite build",
```
Josh-Cena: I'd be surprised if this is needed, because `bun run build` should automatically make `vite` in the `PATH`. This is how npm executors work in general

PR #1841, opened 2025-11-02
fix: add tst url for local auth

reybahl:
fixes the CAS auth for local by using secure-tst link instead of secure link, when in local dev mode


Etherite1 (approved):
LGTM

PR #1847, opened 2025-11-20
feat(calendar): implement calendar toggle + related UI updates

YangaB7:
Replace the previous “expand calendar” behavior with a new dynamic calendar lock toggle for the Worksheet calendar view.

**What it does**
-Introduces a new isCalendarViewLocked state in the store.
-When locked, the calendar always displays a fixed 8am–6pm range.
-When unlocked, the calendar automatically expands or contracts based on the earliest and latest course times.
-Updates related store slices to track and toggle the lock.
-adjusts Worksheet stylesheet for layout alignment with the new toggle.


Josh-Cena:
Why remove the "expand" icon? I think this new feature is entirely unrelated to "expand" right? "Expand" hides the right sidebar (basically, horizontal expansion) and yours fixes the time box (basically, vertical expansion).

Also I feel like the fixed timebox needs to be configurable. For example classes rarely meet before 9am, so the 8am–9am part is unnecessary space; there are also classes that meet as late as 7pm, so you don't want to make them disappear on a fixed display.

YangaB7:
Went ahead and replaced calendar lock button with expandable controls dropdown
- Added eye icon trigger that reveals lock, settings, and expand buttons on hover
- Added modal for customizing locked time range (default 8am-6pm). Accessible through settings button in dropdown

As @Etherite1 mentioned, it may be necessary to have a setting page for the worksheet view. I figured any future settings could also fit neatly into the modal.

Hopefully this does the trick :)

frontend/src/components/Worksheet/CalendarLockSettingsModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+              className={styles.select}
+            >
+              {HOUR_OPTIONS.map((hour) => (
+                <option key={hour} value={hour}>
+                  {formatHour(hour)}
+                </option>
+              ))}
+            </Form.Select>
+          </Form.Group>
+        </div>
```
reybahl: can we use a `CustomSelect` here instead of the native dropdown?

reybahl (commented):
great stuff @YangaB7 - just a couple comments/suggestions

- when i have classes earlier than my start time (e.g., 10am-10pm calendar and 9am class), I can see those classes in my worksheet but the worksheet settings still say 10am-10pm. I think you can either not show that class, or don't allow the start time to be later than any classes or the end time to be earlier than any classes. what do you think?
- use our custom selection dropdown component for start/end times
- use a tooltip instead of title for the beta marker

PR #1855, opened 2025-12-03
feat: map for classes

yav-fi:
## Summary

This PR implements the following feature:

- Adds a _map_ tab option under Worksheet
- In Map view: the map is viewable in place of where the calendar would be in the classic Calendar view
- On the right side, options for settings and hide/view courses + summary just as in the Calendar view are visible. Courses also show locations. The export to calendar option is not present, as the information under Map isn't relevant to the calendar.
- Clicking on a pin on the map for a location briefly highlights courses in the right column that are located at the location
- Hovering over any class in the right column will highlight a pin on the map in orange if there's a relevant class
- any classes without location data or lacking coordinates for the location show an orange warning icon next to their listed locations

## Testing

Run the server (details on CONTRIBUTING.md), go to localhost:3000 and log in -> click worksheet -> click map to test.

### DIFF
--- a/frontend/src/components/Navbar/Navbar.tsx
+++ b/frontend/src/components/Navbar/Navbar.tsx
@@ -79,7 +79,7 @@ export default function CourseTableNavbar() {
       <Navbar
         expanded={navExpanded}
         onToggle={setNavExpanded}
-        expand="lg"
+        expand="md"
         className={clsx(
           'shadow-sm px-3 align-items-start',
           styles.navbar,


frontend/src/components/Worksheet/WorksheetCalendarListItem.tsx
```diff
@@ -9,15 +10,33 @@ import { useStore } from '../../store';
...
+};
+
 export default function WorksheetCalendarListItem({
   listing,
   hidden,
-}: {
-  readonly listing: CatalogListing;
-  readonly hidden: boolean;
-}) {
+  showLocation = false,
```
Josh-Cena: I prefer to not use optional props. We are not building a library so there's no need to be permissive. All these components are used in one or two places so it's better to require explicitness.

Also I wonder if there's a way to not drill down so many props: perhaps a context would be more appropriate here?

yav-fi: I tried to resolve conflicts for WorksheetCalendarList but failed. Could you help me on this? Gotta lock in on my exam for tomorrow...

Josh-Cena: I no longer see a merge conflict, so I assume it's already resolved. Anyway no hurries; we are all busy at this time so we can definitely come back to this later. This already looks really really sick. Thank you!

frontend/src/slices/WorksheetSlice.ts
```diff
@@ -283,7 +284,9 @@ export const useWorksheetEffects = () => {
     exoticWorksheet?.data.season ?? viewedSeason,
     exoticWorksheet ? 0 : viewedWorksheetNumber,
   );
-  setWorksheetInfo(courses, worksheetLoading, worksheetError);
+  useEffect(() => {
```
reybahl: any reason to add this `useEffect` call here? if possible, i think this change should be reverted since this creates an extra render cycle

reybahl (commented):
great work @yav-fi !

just a few suggestions and this PR should be good to go.

reybahl (changes requested):

PR #1867, opened 2026-01-21
misc: seed local friend data on startup

### LINKED ISSUE #1866: Add local seed data for friends
Developers should be able to locally test features involving friend requests. We can add a small local seed script that runs by default on API startup to make it easier to test these functionalities.

aryans-15:
Adds a local dev seed script to create placeholder users (runs automatically on API startup, can be disabled with --no-seed-friends). Note: the script has flags that can be used through manual runs. Closes #1866


api/start.sh
```diff
@@ -12,6 +12,7 @@ ENV=""
 OVERWRITE=false
 FERRY_SEED=false
 OFFLINE=false 
+SEED_FRIENDS=true
```
reybahl: instead of defaulting to true, we could do something like the following:

* Default `SEED_FRIENDS` to false.
* If `--ferry_seed` is active, automatically set `SEED_FRIENDS` to true (unless explicitly disabled), so that when you re-seed ferry it won't wipe friends entirely
* Add a --seed-friends flag so you can seed friends without seeding ferry

what we get with seeding friends by default is resetting the friends data across server starts which might not be intended.

api/drizzle/seed-friends.ts
```diff
@@ -0,0 +1,128 @@
...
+import { drizzle } from 'drizzle-orm/postgres-js';
+import postgres from 'postgres';
+import { studentBluebookSettings, studentFriendRequests } from './schema.js';
+
+const { values } = parseArgs({
+  options: {
+    count: { type: 'string', default: '5' },
+    requests: { type: 'string', default: '0' },
+    prefix: { type: 'string', default: 'test' },
+    clean: { type: 'boolean', default: false },
```
reybahl: Should we default --clean to true (or just always clean)? Without clean, running this multiple times piles up more data which could messy. Up to you.

PR #1882, opened 2026-02-06
feat: prompt latest offering for historical course adds

### LINKED ISSUE #1787: Adding a course from the past should be able to add its current version if popssible
Currently, we issue a warning when you try to add a historical offering: "This will add to a worksheet of a semester that has already ended". While this is enough for the user's information, maybe we could go one step further for their experience: for historical courses, we should get its latest offering and see if it's a "current" one (in a future semester). If so, we could prompt "do you want to add the latest one instead, or do you still want to add this one". There's not even such a warning on the catalog itself (the warning is only for the modal), so this feature is even more necessary!

aryans-15:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Summary

When a user adds a past course, we display a modal asking if they meant to add the historical or the latest version.

## Test plan (required)

Popup modal looks like this
image

## Related issues/PRs

Closes #1787

### DIFF
--- a/frontend/src/components/Worksheet/WorksheetToggleButton.tsx
+++ b/frontend/src/components/Worksheet/WorksheetToggleButton.tsx
@@ -190,7 +190,7 @@ function WorksheetToggleButton({
             query: LatestCurrentOfferingDocument,
             variables: {
               sameCourseId,
-              seasonCodes: seasons,
+              seasonCodes: seasons as string[],
             },
           });
           const [latestCourse] = data.courses;


frontend/src/components/Worksheet/WorksheetToggleButton.tsx
```diff
@@ -129,11 +177,54 @@ function WorksheetToggleButton({
...
+                targetCrn = latestListing.crn;
+                targetWorksheetNumber = getRelevantWorksheetNumber(
+                  latestCourse.season_code,
+                );
+              }
+            }
+          }
+        } catch {
+          // If lookup fails, fall back to adding the selected listing
+        }
```
reybahl: can we add some error logging via Sentry? would be helpful in debugging in case some issue did happen

Issue #1912, opened 2026-03-21
Public profile pages


reybahl:
- Public (or shared) profile experience: dedicated route(s) or modal, not only “my” profile.
- Friend search: clickable names (or rows) to open that person’s profile / limited card.
- Visibility model: define who can see what (strangers vs friends vs self); API returns only allowed fields; UI matches.
- Align with #1911 where users set preferences (e.g. preferred name)

PR #1916, opened 2026-03-24
feat: profile settings, public profiles, friend search

### LINKED ISSUE #1911: Profile customization & privacy
- Customize how you appear: e.g., preferred first/last name
- Fill in names when they’re missing in the DB.
- Revoke evaluations access

aryans-15:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Summary

Closes #1911, #1912 

## Release Notes

* **New Features**
  * Added a public user profile page accessible via direct link with privacy-controlled information display.
  * Added profile settings tab with visibility controls for name, email, and other profile fields.
  * Replaced local friend search with live remote profile search functionality.
  * Added ability to revoke evaluation access from user settings.

* **Style**
  * Updated form element styling with smooth transitions.
  * Added comprehensive styling for new user profile and settings layouts.


reybahl:
also: when there’s no preferred override, the resolved name only shows as a placeholder, so focusing the field looks empty and it feels like you have to retype your name. might be better to prefill the input with the same resolved string you’d use as placeholder, and on save send null when the trimmed value matches the legal name so we don’t store a redundant override.

like this:
Screenshot 2026-04-05 at 15 11 02

reybahl (commented):
one more thing, i feel like we don't need this custom screen and instead we could just show page not found, wdyt?

image

frontend/src/pages/UserProfile.tsx
reybahl, line 1: minor ux: on `/u/:netId`, the header shows the NetID from the URL during loading, so a profile page screen flashes before 404 when the page isn’t actually available. Think it might be better to hide the identity header until load finishes or use a generic loading state

frontend/src/pages/Profile.tsx
```diff
@@ -89,6 +181,25 @@ function Profile() {
     if (!friendRequests) void friendReqRefresh();
   }, [friendRequests, friendReqRefresh]);
 
+  useEffect(() => {
+    void friendReqRefresh();
+  }, [friendReqRefresh]);
```
reybahl: we could consolidate all three of these `useEffect`s into one:

```
useEffect(() => {
    if (!currentUser) return;
    if (activeTabKey !== 'overview') return;
    void friendRefresh();
    void friendReqRefresh();
  }, [activeTabKey, currentUser, friendRefresh, friendReqRefresh]);
```

reybahl: since we only need this for overview, and we should refresh every time we open overview

reybahl:
also can you make the friend remove button show a confirmation toast like the previous one does?

frontend/src/components/Worksheet/AddFriendDropdown.tsx
```diff
@@ -165,46 +155,53 @@ function AddFriendDropdownDesktop({
...
+                label: profile.displayName
+                  ? `${profile.displayName} (${profile.netId})`
+                  : profile.netId,
+                type: 'searchResult',
+              }),
+            ) ?? [];
+        setSearchResults(nextResults);
+        setIsLoading(false);
+      });
+    }, 180);
```
reybahl, line 206: `setIsLoading(false)` after await won’t run if the friend action rejects, can we use try/finally in handler?

reybahl:
aside: maybe we should (in a follow up pr perhaps) add a profile preview page? where you can see what your profile looks like for someone else? and you can toggle between friend and public view. unless its small enough to add to this PR, probably worth adding as a follow up pr

reybahl (commented):
all looks good to me 👍  before we merge, i think it would make sense to default the email visibility to `friends` instead of `self`, wdyt? right now looks a little odd since for friends view i can see all the information except email

reybahl (approved):
🔥

PR #1957, opened 2026-04-19
feat: improve search relevance with aliases and typo tolerance

YangaB7:
## Improve search matching and ranking

Search is now more flexible and results are ordered more intuitively.

Adds subject full-name matching so queries like `computer science` return CPSC courses, introduces conservative typo tolerance (including common transposition mistakes), and adds a small alias layer for shorthand like `cs` and `polisci`.

Also adds a lightweight match-quality signal so stronger matches appear above weaker ones. Exact and direct matches are prioritized, while fuzzy matches still work but don’t dominate results.

## To test, try:

- `cs`, `cs 201`  
- `computer science`  
- `algorthms`, `intrduction`  
- `orgo`, `polisci`

## Summary by CodeRabbit

* **New Features**
  * Enhanced search with graded match-quality scoring, improving ranking by relevance during text searches
  * Fuzzy/typo-tolerant matching (with safeguards for short tokens) and alias-aware token expansion for broader discovery

* **Tests**
  * Added comprehensive tests covering fuzzy matching, alias behavior, exact-vs-fuzzy distinctions, multi-token AND semantics, and match-quality ranking

### DIFF
--- a/frontend/src/search/SearchBootstrap.tsx
+++ b/frontend/src/search/SearchBootstrap.tsx
@@ -15,7 +15,7 @@ import {
   emptyFilters,
   SEARCH_FILTER_KEYS,
 } from './searchConstants';
-import { matchesSearchText } from './searchTextMatch';
+import { searchMatchQuality } from './searchTextMatch';
 import type {
   BooleanAttributes,
   FilterHandle,
@@ -425,6 +425,7 @@ export function SearchBootstrap({
         return [...data.values()];
       });
 
+      const qualityMap = new Map<CatalogListing, number>();
       const filtered = listings.filter((listing: CatalogListing) => {
         // For empty bounds, don't apply filters at all to include no ratings
         if (overallBounds.isNonEmpty) {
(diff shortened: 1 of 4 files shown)


frontend/src/search/SearchBootstrap.tsx
```diff
@@ -425,6 +425,7 @@ export function SearchBootstrap({
         return [...data.values()];
       });
 
+      const qualityMap = new Map<CatalogListing, number>();
```
reybahl, line 428: NIT: when there are no search tokens, `searchMatchQuality` still returns a max score for every course, so we fill `qualityMap` for the whole list even though it doesn’t change sort order. For efficiency we could short circuit - if `processedSearchTextParam.length === 0`, don’t build the map and pass undefined to sortCourses (or only attach scores when length > 0)

frontend/src/search/searchTextMatch.ts
```diff
@@ -2,43 +2,171 @@ import type { CatalogListing } from '../queries/api';
...
+
+  return false;
+}
+
+// Common student shorthand → tokens to also try
+const searchAliases: { [key: string]: readonly string[] } = {
+  cs: ['cpsc'],
+  orgo: ['organic'],
+  polisci: ['plsc'],
+};
```
reybahl, line 63: might be worth noting this somewhere, maybe add a new section to `maintenance.md`? i think might be good for future contributors / maintainers to know that these search aliases can be updated, esp if they change in the future

reybahl: maintenance.md is in the `/docs` folder btw

reybahl (commented):
just making those two comments, code looks good otherwise @YangaB7 🙌

